PR #86, opened 2026-04-13
Move app navigation to the side in landscape

surma:
## Summary

This switches the main app navigation from a bottom bar to a side rail when the device is in landscape orientation.

Behavior after this change:

- portrait keeps the existing bottom navigation bar
- landscape uses a `NavigationRail`
- desktop fullscreen still hides navigation entirely
- the existing pager content and screen routing stay the same
- bottom-nav drag-to-reorder behavior is preserved in portrait mode

The landscape rail uses the current screen order, so it stays consistent with the user’s saved navigation order.

## Why

In landscape, vertical space is more valuable and horizontal space is easier to spare.

Moving navigation to the side gives terminal/desktop/content screens more usable height without changing the portrait experience.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -132,6 +132,7 @@ dependencies {
     implementation(platform(libs.compose.bom))
     implementation(libs.compose.ui)
     implementation(libs.compose.material3)
+    implementation(libs.compose.material3.window.size)
     implementation(libs.compose.material.icons)
     implementation(libs.compose.ui.tooling.preview)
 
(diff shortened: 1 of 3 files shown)


GlassOnTin:
Pushed 6753e6 to restore the `#78` locale/maxLines comment and the instant-scroll note that were dropped when the bottom-bar block moved inside the `!useSideNavigation` branch — the WHY is worth keeping with the code.

One discussion point, not a blocker: `Configuration.ORIENTATION_LANDSCAPE` triggers the rail on any landscape phone, including narrow devices where the rail eats more width than it saves. `WindowSizeClass` (width ≥ Medium) is the Material3-recommended signal and would also opt tablets in portrait into the rail. Worth considering whether that fits Haven's device profile better, or whether orientation is deliberately the right knob here.

surma:
Good shout. Implemented that!

PR #111, opened 2026-04-25
Show friendly error when password auth fails (Auth cancel → check credentials)

jensrot:
When a password connect attempt fails, JSch throws 'Auth cancel for methods publickey,password' — internal jargon that reads as if the user cancelled something. This replaces it with 'Authentication failed — check username and password' for the keyOnly=false path.


GlassOnTin:
Thanks @jensrot — clean fix and the refactor (extracting `isAuthMessage` from the `keyOnly && …` gate) is exactly the right shape for the new branch you added. Going to merge.

I noticed two sister sites with the same pattern that surface the same JSch jargon:
- Mosh path at `ConnectionsViewModel.kt:1779`
- Eternal Terminal path at `ConnectionsViewModel.kt:1650`

I'll mirror your pattern to both in a follow-up commit on main so the friendlier message lands across all three SSH-based transports — your authorship on the SSH fix stays intact.

Both ship in v5.24.44.

Issue #92, opened 2026-04-14
Mosh still Freezing after Commands


gitcodeerrors:
The freezing was also in [https://github.com/GlassOnTin/Haven/issues/73](https://github.com/GlassOnTin/Haven/issues/73)
I have opened a new issue, since the other issues from there have been fixed!

> Do arrow keys navigate in the Mutt message list? (primary fix)

With version 4.5.2 the Arrow Keys works great.

> Do the "cursor below prompt on login"

Also fixed! Never happen after fix.
Thank you.

Still open
> "freezes after a few empty Returns" symptoms recur?
> Both might be downstream of the broken arrow keys and go away with this fix, 
> or they might be a second rendering-semantic bug I'll need to chase separately.
> Either answer is useful.

Yes ist still happens.
I can type ls or other commands but after a while if I am trying to logout with
exit
logout
or STRG+D
it freezes.
The strange thing in is I can type exit but then Return does not work and
I have to terminate it on the connection page.

GlassOnTin:
Thanks for the detailed retest @gitcodeerrors — confirming the arrow-key and cursor-below-prompt fixes work in v4.5.2 is useful, and the residual *"Return doesn't fire when I try to log out"* symptom is a real one I want to chase.

The pattern you describe — characters type fine for a while, then a specific moment where the Enter key stops being delivered to the shell — doesn't match a generic input-handling bug because plain text is still going through. A few hypotheses worth narrowing down before I touch any code:

1. **The shell exited but the client doesn't realise.** When `exit` runs, the remote shell terminates and its PTY child closes. mosh-server doesn't tear the session down immediately — it keeps the framebuffer intact. From your end it looks like *"I typed `exit`, the next Return goes nowhere"* because there's no longer a process listening for stdin on the other side. If this is the cause, the fix is to detect PTY EOF on the server side and surface a "session ended" state to the client so the connection card knows to drop instead of leaving you stuck on a dead screen.

2. **A receive-pipeline state mismatch carrying over from the #73 work.** The base-mismatch tracking I added in v4.5.2 (`ReceiveState.tryAdvance`) intentionally drops diffs whose `oldNum` doesn't match the client's current state. There's a narrow window where the server can compute a diff against a state the client has just advanced past, which would silently drop the "exit echoed / shell prompt returns" output. I'd expect that to look like a frozen prompt rather than as a Return-key failure though.

[...]

PR #365, opened 2026-07-10
feat: mosh attach + per-profile timeout + kill/reconnect toggle

Biotoza:
## Problem
1. Mosh sessions disconnect after 8s of silence — too fast for mobile networks
2. No way to reconnect to an existing mosh-server (key is lost on disconnect)
3. No per-profile timeout control
4. No way to kill old mosh-server processes when starting fresh

### New profile fields
| Field | Type | Default | Description |
|---|---|---|---|
| moshSessionTimeoutSec | Int | 120 | Session-dead timeout in seconds (0 = library default) |
| moshReconnectToExisting | Boolean | true | Reconnect with saved key vs. kill+new |
| moshShowTimeoutSlider | Boolean | false | UI flag for per-profile slider |

### mosh attach (reconnectTerminalSession)
New method in MoshSessionManager re-establishes UDP to the SAME
mosh-server using the saved (ip, port, key). SSP re-syncs state.

### Per-profile timeout
sessionDeadMs flows: ConnectionProfile → MoshSessionManager.SessionState
→ MoshSession → MoshTransport constructor → effectiveSessionDeadMs
(replaces hardcoded SESSION_DEAD_MS in sendLoop).

### Kill-on-reconnect
When moshReconnectToExisting=false, SSH execs pkill mosh-server
before starting a new one — no zombie accumulation.

### DB migration
76→77 adds three columns to connection_profiles.

## Related
Companion PR in GlassOnTin/ssp-transport#1 (relaxed timeouts + sessionDeadMs param).


GlassOnTin (changes requested):
Thanks for this — the diagnosis is spot on, and reconnect-with-same-key is exactly the right mosh semantic. Today an 8-second UDP outage fires `onDisconnected`, the status flips to DISCONNECTED, and `TerminalViewModel` removes the tab, orphaning a perfectly alive mosh-server. Worth fixing.

Tracing the paths as written, though, the new plumbing never actually executes, and there are a couple of lifecycle bugs to sort before this can land:

**1. No caller.** Nothing invokes `reconnectTerminalSession()` — no auto-reconnect trigger, no UI affordance. And the tab has already been removed by the time status hits DISCONNECTED (`TerminalViewModel.kt:1018` filters tabs to CONNECTED sessions), so there's nothing left to reattach to. For this to do anything, the disconnect path needs to keep the tab alive (e.g. a "reconnecting" overlay) and drive the reconnect itself — ideally auto-retry with backoff from the `onDisconnected` callback.

**2. The `isReadyForTerminal` DISCONNECTED branch is unreachable in the scenario it targets.** It still requires `moshSession == null`, but the disconnect path never clears `moshSession`, so after a UDP timeout the check stays false. (Only `TerminalViewModel:1179` calls this for mosh, and that loop iterates CONNECTED-only sessions anyway.)

**3. Old transport leaked.** `reconnectTerminalSession` overwrites `session.moshSession` without closing the previous one — the old UDP socket and threads are never released, and it doesn't guard against being called while a live session is still attached.

**4. The new session is never started.** The method returns the `MoshSession` without calling `start()`, and status is set to CONNECTED synchronously before any SSP re-sync happens. If the server is gone you'd show CONNECTED for up to SESSION_DEAD_MS.

**5. `cleanExit` isn't actually used.** The PR description says clean exits don't reconnect, but `cleanExit` isn't stored anywhere — eligibility only checks DISCONNECTED, so a `logout` would still be "eligible" and churn against a dead port. Store it on the session state and gate reconnect on it.

**6. Minor:** the reconnect session drops `verboseBuffer`, so reconnected sessions lose verbose transport logging.

On the companion timeout PR (GlassOnTin/ssp-transport#1): raising `SESSION_DEAD_MS` to 120s is the right call — arguably real mosh never gives up client-side at all. But `NETWORK_STALL_MS` 6s→45s is a regression: that's the socket rebind that recovers Wi-Fi→LTE roaming, and delaying it means up to 45s of dead air on an interface switch — the one case mosh is famous for handling near-instantly. I'd keep the rebind at ~6s (or rebind periodically) and raise only the death threshold. I've left a note there too.

Happy to take this over the line if you'd rather — the shape I'd aim for: keep the tab on DISCONNECTED with a reconnecting indicator, auto-invoke the reconnect with backoff, close the old transport first, gate on `cleanExit`, and carry `verboseBuffer` across. If you want to iterate on it yourself, points 1–4 are the blocking ones.

GlassOnTin:
Hold off on further commits for a moment — I've dug into the transport while reviewing this and found a cleaner root-cause fix that I'm landing now; I don't want you sinking more time into the timeout/knob direction.

The key discovery: mosh-server doesn't go silent when the shell exits — it *announces* shutdown by sending a state numbered -1 (upstream `Transport::start_shutdown()`). Our Kotlin client never handled that, which is why the transport had to guess "dead shell vs dead network" from silence (`SESSION_DEAD_MS`), and guessed wrong on network outages. Once the client detects the announced shutdown, silence only ever means "network away" — so the transport can simply retry forever with the same key (true mosh semantics), rebinding its socket periodically for roaming. Reconnect-to-existing becomes inherent; there's no timeout left to configure.

That makes the v2 additions unnecessary, and two of them are problematic on their own:

- The migration bakes `moshSessionTimeoutSec` and `moshShowTimeoutSlider` into the profile schema — the first becomes meaningless with no client-side timeout, and the second is UI state, which doesn't belong in the DB.
- `pkill -U $(id -u) mosh-server` on the remote host kills *all* the user's mosh-servers — including live sessions from other devices. That's a destructive default for a convenience toggle.

Also a compile note: `MoshSession(sessionDeadMs = …)` passes a parameter the pinned ssp-transport submodule doesn't have, so the PR as pushed won't build without a submodule bump.

Full write-up coming on this thread once the fix is verified on-device. Your diagnosis of the 8-second kill was what prompted the dig — thank you; I'll credit you on the commit.

GlassOnTin:
Landed on main as 04d0df79 (Haven) + GlassOnTin/ssp-transport@34c059b (transport), with you as co-author on both. Closing this PR — here's the full story.

## Root cause

The transport couldn't distinguish "shell exited" from "network away": both looked like silence, so it guessed via an 8-second timer (`SESSION_DEAD_MS`) and reported the timeout as a *clean* exit. Your diagnosis of the symptom was exactly right — an 8s Wi-Fi blip killed a session that mosh-server was happily keeping alive.

The missing piece: mosh-server doesn't go silent when the shell exits. It **announces** shutdown by sending a state numbered -1 (uint64 max — upstream `Transport::start_shutdown()`). Our Kotlin client never handled that sentinel, which is the only reason the silence heuristic existed at all.

## The fix

- **Detect the announced shutdown** (`newNum == -1`) → close with `cleanExit=true`. Verified against the real mosh-server binary in a new E2E test (`RealServerE2ETest`, gated on `MOSH_SERVER_E2E=1`): echo round-trips, `exit` produces the announcement, the server quits on its own ~3s later without our ack.
- **No client-side give-up on silence.** The transport now retries indefinitely with the same key — SSP re-syncs whenever connectivity returns, matching upstream mosh. "Reconnect to existing session" is therefore inherent; no reconnect machinery, no timeout knob, no schema migration needed.
- **Periodic socket rebind** (every 6s of continued silence, was one-shot) so an interface change during a long outage still recovers.
- **UI**: the "Disconnected — closing in Ns" countdown is now a "No server contact for Ns — retrying" count-up (all 12 locales).

## Device verification (Pixel, release build)

Before (v5.68.38 as shipped): froze mosh-server with `SIGSTOP` → session DISCONNECTED after ~8s, tab gone, server orphaned.

After: froze the server **51 seconds** → tab alive, banner counting up ("No server contact for 42s — retrying"), logcat showing rebinds every 6s → `SIGCONT` → `pwd` round-tripped immediately. Then `exit` → logcat `Server announced shutdown (newNum=-1), closing` → `clean=true` → tab closed and mosh-server exited host-side.

## Why not the v2 approach

With no client-side timeout there's nothing for `moshSessionTimeoutSec` to configure, and `moshShowTimeoutSlider` is UI state, which shouldn't live in the profile schema. The `pkill -U $(id -u) mosh-server` path was also dangerous — it kills the user's mosh-servers from *all* their devices, not just the orphan. None of it is needed now.

One real gap does remain, if you're looking for a follow-up with teeth: **reattach after app restart**. The (ip, port, key) tuple only lives in memory, so a killed/restarted Haven can't rejoin a still-alive mosh-server. Persisting the key safely (it's a session credential) and re-running discovery on restore is a genuinely useful feature — happy to review a design for it if you want to take it.

Thanks again for the report and the push on this — the 8s kill had been mislabelled "clean exit" since #92 and nobody had chased it to the protocol level.

Issue #356, opened 2026-07-08
CI: armv7 AAB packaging intermittently OOMs (bundletool 'Java heap space') — bump Gradle heap


GlassOnTin:
## Symptom

The release workflow's **armv7 AAB packaging** intermittently fails with an OOM:

```
Execution failed for task ':app:packageArmv7ReleaseBundle'.
> A failure occurred while executing ...PackageBundleTask$BundleToolWorkAction
   > Java heap space
```

Seen on the v5.68.24 release run (the `build (armv7, Armv7)` job); a `--failed` rerun of the same commit packaged cleanly, so it's non-deterministic, not a code fault. `verify-ci`, `test`, and the arm64/x64 build jobs all passed — only armv7 packaging tipped over, and it's the last ABI in the matrix, so heap accumulated in the daemon is the likely trigger.

## Cost

Each hit wastes a full release cycle: the run reaches the ABI builds (~15 min in), fails one job, and needs a manual `gh run rerun --failed` before the signing gate + publish can proceed. It has cost a rerun on at least one release already.

## Proposed fix

Bump the Gradle daemon heap in `gradle.properties`:

```diff
-org.gradle.jvmargs=-Xmx4096m -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx6144m -Dfile.encoding=UTF-8
```

bundletool's packaging runs in the Gradle daemon; 4 GB is tight when it follows two prior ABI packagings in the same daemon. GitHub's `ubuntu-latest` has ~16 GB RAM, so 6 GB is comfortably within budget for CI (and for local release builds). Alternatively raise it only for the release job via `GRADLE_OPTS`, but the global bump is simpler and harmless.

## Why not fixed already

Deliberately not applied speculatively: the runner OOM can't be reproduced locally to confirm 6 GB actually resolves it, and changing CI config on an unverified hypothesis is the wrong call. The intended path is **apply the bump, then watch the next few release runs** — if armv7 packaging stops OOMing, keep it; if it recurs, the accumulation theory is wrong and it needs worker-process isolation or a per-ABI daemon instead.

Filed while shipping v5.68.24 (which hit this).

GlassOnTin:
Applied the heap bump (`org.gradle.jvmargs` 4096m → 6144m) in v5.68.26 (`3a06d987`).

[...]

Issue #164, opened 2026-05-18
Mosh does not work when tunnel is enabled


Senkoi:
When the built-in tunnel(wireguard) is enabled in Haven:

* SSH works normally
* Mosh cannot connect

If the built-in tunnel is disabled and an external VPN application is used instead:

* SSH works normally
* Mosh also works normally

This issue only occurs when using the built-in tunnel feature.

GlassOnTin:
Investigated this for v5.39 prep but it's slipping to v5.40 — the fix is larger than a normal bug pass.

Root cause: Haven's `TunnelResolver` only exposes a TCP surface today. `dial()` returns a `TunneledConnection` wrapping a TCP `Socket`. Mosh's `MoshConnection` creates a raw `java.net.DatagramSocket()` directly, which bypasses the WireGuard tunnel entirely — Mosh's UDP packets leave the device on the default route. SSH works because it's TCP and rides the tunnel; Mosh sits beside it.

The fix needs:
1. A new `TunneledDatagramSocket` interface in `core/tunnel/` with `send` / `receive` primitives that go through `wireguard-go`'s userspace UDP packet pipe.
2. `TunnelResolver.dialUdp(profile, host, port)` returning that, or `null` to fall through to the plain UDP path.
3. `MoshConnection` refactor to accept an optional socket provider — defaulting to the current `DatagramSocket()` behaviour when no tunnel is set.

Sized at ~1 week of focused work plus on-device validation against a Mosh-only-through-WG server. Queued for v5.40 in the project plan; in the meantime, the workaround is to use an external WireGuard app (which routes UDP at the OS level) rather than Haven's built-in tunnel.

Thanks for filing — the root cause was useful and would have been hard to catch without your TCP-vs-UDP split observation.

GlassOnTin:
v5.40.0 is up: https://github.com/GlassHaven/Haven/releases/tag/v5.40.0

What landed: a UDP surface across the gomobile bridge (the WireGuard / Tailscale Go side already supported UDP via gVisor netstack / tsnet, it just wasn't exposed to Kotlin). `TunnelResolver` gained `dialUdp` + `udpSocketSupplier`; `MoshConnection` now takes a pluggable UDP socket factory; non-tunnelled Mosh profiles fall through to the same raw `DatagramSocket` they used before, so direct sessions are unchanged.

[...]

PR #371, opened 2026-07-10
feat: mosh session key persistence — reconnect after app restart

Biotoza:
## Problem
When Haven is killed/restarted, the mosh session key (ip, port, key tuple) is lost — it only lives in memory. The server-side mosh-server is still alive (7-day TTL), but the client has no way to rejoin it.

## Solution
Persist the session credential to ConnectionProfile after a successful mosh-server bootstrap, and use it on the next connect to skip the SSH+mosh-server-new bootstrap entirely.

### How it works
1. After `finishMoshConnect()` parses MOSH CONNECT, the (ip, port, key) is saved to the profile via `repository.save()`
2. On next connect, `connectMosh()` checks for a saved key + `moshReconnectToExisting == true`
3. If found: calls `finishMoshReconnect()` directly — no SSH, no new mosh-server. Just opens UDP with the saved key. SSP re-syncs state.
4. If the user chooses "New session" (`forceNewSession = true`): runs `pkill -U \1000 mosh-server` to clean zombies, then normal SSH bootstrap

### New profile fields
| Field | Type | Purpose |
|---|---|---|
| savedMoshKey | String? | Session key from last bootstrap |
| savedMoshPort | Int? | UDP port |
| savedMoshServerIp | String? | Server IP |

### What was removed (cleanup from old PR #365)
- `moshSessionTimeoutSec` — transport now retries forever (GlassOnTin/ssp-transport@34c059b)
- `moshShowTimeoutSlider` — UI config doesn't belong in DB schema

### DB migration
77→78: adds three saved-mosh columns. Old timeout/slider columns left in SQLite (no DROP COLUMN on older versions) but Room no longer maps them.

## Related
- GlassOnTin/ssp-transport#1 (root cause fix: detect mosh-server shutdown announcement, retry forever)
- GlassHaven/Haven#365 (first attempt — timeout approach, superseded)


Biotoza:
Hey, anything else needed from my side on this one?

GlassOnTin (changes requested):
Sorry for the slow review — the ball was entirely on my side. The direction is exactly right: this implements the review points from #365 faithfully, and the TerminalViewModel auto-reconnect machinery (keep the tab, overlay + backoff, in-place transport swap preserving the emulator) is the design I hoped for. Thank you for sticking with it through three rounds.

That said, the branch as pushed can't build, so I suspect a commit or hunk went missing before the push. Full pass below.

## Blocking — doesn't compile

1. **`finishMoshReconnect(...)` is called but never defined.** `connectMosh`'s saved-key fast path calls it, but the function doesn't exist anywhere on `feature/mosh-key-persist` — unresolved reference. Looks like the commit containing it didn't make it into the push.
2. **`private const val MOSH_RECONNECT_MAX_DELAY = 30_000L`** sits in the `TerminalViewModel` class body (~line 937). `const val` is only legal at top level or inside an `object`/`companion object` — this won't compile. Plain `private val` is fine here.

Before the next push, please run the local gates so we catch this class of thing early:
`./gradlew :feature:connections:compileDebugKotlin :feature:terminal:compileDebugKotlin :core:mosh:compileDebugKotlin testDebugUnitTest`

## Blocking — must fix before merge

3. **`gradle.properties` is replaced wholesale with a local config.** It drops `android.useAndroidX=true` (build breaks), `org.gradle.caching`/`parallel`, and the 6144m heap that fixed the armv7 AAB OOM (#356). Please revert that file completely — it shouldn't be in this PR.
4. **`savedMoshKey` is persisted in plaintext.** Every other secret on `ConnectionProfile` (sshPassword, vncPassword, spaKey, …) goes through `CredentialEncryption` in `ConnectionRepository.encryptPasswords()`/`decryptPasswords()`. The mosh key is a live session credential — whoever holds it owns the terminal session for up to 7 days — so it must be added to both lists (and to the plaintext-upgrade sweep in the repository `init`).
5. **The `pkill` never fires.** `"pkill -U \\$(id -u) mosh-server 2>/dev/null; echo done"` sends the literal bytes `\$(id -u)` to the remote shell — the backslash suppresses the command substitution, so pkill gets the literal string `$(id -u)` as its user argument and matches nothing. Drop the backslash. (Good news: this also means the scary path has never actually killed anything yet.)
6. **`forceNewSession` is dead code.** Nothing ever passes `true` — `connectMosh` declares it, doesn't forward it to `finishMoshConnect`, and the "New session" dialog choice from the PR description isn't wired to any UI. Either wire it end-to-end or drop the parameter + kill block from this PR and do it as a follow-up. If you wire it: note that `pkill -U` will also kill the mosh-servers behind this profile's *other live Haven tabs* on the same account, not just zombies — the dialog copy needs to say that, or the kill needs a narrower target.

## Blocking — stale-key behaviour

7. **The kdoc promises "wiped when the server does a clean exit", but nothing wipes it.** There's no `repository.save(... savedMoshKey = null ...)` on the `cleanExit = true` path. After the user types `exit`, the next connect will re-attach to a dead port.
8. **A dead/rebooted server can't fail the fast path.** UDP is connectionless, so `finishMoshReconnect` won't throw for a gone server — the `catch` around the fast path only covers synchronous setup failures. Combined with (7), a stale tuple leaves the tab stuck in "Reconnecting…" with no automatic escape. Suggestion: bound the re-attach with a liveness window — if no server datagram arrives within N seconds, tear down, wipe the saved tuple, and fall through to the full SSH bootstrap automatically. That also covers the server-IP-changed case (DHCP/roaming; see #376) without user intervention.

## Rebase + migration

9. The PR is conflicting and ~58 commits behind; main is at DB **76** (it moves fast here, so re-check at rebase time). Please collapse `MIGRATION_76_77` + `MIGRATION_77_78` into a single migration at the right number — neither has shipped, so there's nothing to step through. Also drop the kdoc about removing `moshSessionTimeoutSec`/`moshShowTimeoutSlider` columns: those never landed on main, so for everyone but your test devices the story is just "adds four columns". And commit the exported Room schema JSON (`core/data/schemas/sh.haven.core.data.db.HavenDatabase/<n>.json`) that the version bump generates.
10. **Drop the `sessionDeadMs` threading** through `MoshSession`/`MoshSessionManager`. It's always `null`, and its kdoc references `ConnectionProfile.moshSessionTimeoutSec` — a field this PR's own description says was removed. It's a leftover from the abandoned timeout approach; the transport default is the behaviour we want.

## Nits

- `moshReconnectBackoff` comment says "(seconds)" but the values are milliseconds.
- `SessionState.cleanExit` kdoc still references `SESSION_DEAD_MS` by name; say "the transport's session-dead timer" instead so it survives renames.
- `reconnectTerminalSession` flips the state to CONNECTED unconditionally — worth a one-line comment that "connected" is optimistic under UDP and the transport's next disconnect callback is the real verdict (it feeds the retry loop in (8)).
- One saved tuple per profile means two tabs to the same host race on `finishMoshConnect`'s save and only the last survives an app restart. Fine as a v1 constraint, but worth a sentence in the field's kdoc.

Once it compiles, is rebased, and 3–6 are addressed, I'll kick CI and device-test the outage → auto-reconnect and app-restart → re-attach flows on real hardware before merging. The structure is right — this is close.

GlassOnTin:
Two corrections to premises in the PR description, from the #365 round — they change points 6–8 above:

- **There is no 7-day TTL.** We established during #365 that a default `mosh-server` has *no* idle timeout at all — it lives until it exits or the host reboots. That cuts both ways: re-attach works even weeks later (good), but stale tuples never age out on their own and orphaned servers accumulate forever, which makes the wipe-on-clean-exit and wipe-on-failed-liveness-probe behaviour from points 7–8 load-bearing, not nice-to-have. Please also fix the "7 days" claims in the kdoc/comments.
- **Remote `pkill mosh-server` was explicitly rejected in the #365 review** (it kills sessions belonging to the user's other devices — and, as noted above, this profile's own other tabs). Gating it behind an explicit "New session" choice doesn't really change that calculus, since users won't connect the dialog button to "my laptop's session just died". Let's drop the kill block from this PR entirely; if zombie cleanup turns out to matter in practice, it deserves its own issue with a per-session target (e.g. remember the specific server's PID from `MOSH CONNECT`-time `$!` capture in a wrapper, or kill by exact key match) rather than a UID-wide sweep.

GlassOnTin:
Rather than send you through another round for what was mostly plumbing into code that shifted under you (main refactored the Mosh/ET bootstrap this week), I've taken the liberty of finishing it on this branch — `maintainerCanModify` was on. You're co-authored on the commit; thanks for driving this to a working design.

What landed (`4aeb1753`), relative to your last push:

- **Kept**: your core design — persist the (ip, port, key) tuple on the profile at bootstrap, re-attach on the next connect without SSH, `cleanExit` tracking on the session state, the per-profile toggle (now surfaced in the connection editor's MOSH section, 12 locales, and on the `update_connection` MCP tool).
- **Added the missing `finishMoshReconnect`** — it mirrors the tail of `finishMoshConnect` minus the SSH client, and deliberately skips `initialCommand`: on re-attach the server-side shell is exactly where you left it, so typing a `tmux attach` into it would corrupt whatever is running.
- **Encrypted the key at rest** — `savedMoshKey` joins the `CredentialEncryption` field set like every other stored secret.
- **Liveness probe instead of hope**: UDP can't distinguish dead from silent, so a new `MoshSession.serverContacted` flow flips on the first genuine server bytes (a fresh transport always triggers a full repaint). No bytes within 10s of transport start → wipe the tuple, fall back to the full SSH bootstrap automatically. Clean exit wipes eagerly, key-matched so an old tab exiting can't clobber a newer bootstrap's tuple.
- **Dropped the in-tab backoff/reconnect machinery**: since `04d0df79` (the #365 landing) the transport itself retries outages forever and roams — `onDisconnected` only fires for announced shutdown or fatal local errors, so the mid-session case your TerminalViewModel loop targeted no longer reaches the app layer. The app-restart path was the real gap, and that's what shipped.
- **Dropped `pkill` / `forceNewSession`** per the review — the toggle plus automatic fallback covers "I want a fresh session" without UID-wide kills.
- Re-attach is skipped while the profile already has a live mosh tab (two transports on one key fight over the server), and tunneled-UDP profiles (#164) resolve the same socket provider on re-attach.

Full unit suite is green; I'll device-test the restart → re-attach and stale-tuple → fallback flows on real hardware next and merge once that passes.

GlassOnTin:
## Stop — the premise doesn't hold. mosh-server won't accept a second client.

I finished the implementation on this branch and device-tested it, and the tests turned up something that kills the approach outright: **a mosh-server will not talk to a new client process, even with the correct key, port and IP.** The session is bound to the *first* client instance; a fresh client's packets are ignored entirely. That's upstream mosh behaviour, not a Haven bug — and it means "the server is still alive after an app restart, so we can re-attach to it" cannot work.

### Reproduction (stock mosh, no Haven involved)

```
$ mosh-server new -s -c 256 -l LANG=en_US.UTF-8
MOSH CONNECT 60008 <KEY>

# client A — the first client this server ever sees
$ MOSH_KEY=<KEY> mosh-client 192.168.0.180 60008
ian@msi-z790:~$ echo SESSION-STATE-MARKER
SESSION-STATE-MARKER            # works

# kill client A. Server is still running and still bound:
$ ss -unlp | grep 60008
UNCONN 0 0 192.168.0.180:60008 ... users:(("mosh-server",pid=...))

# client B — fresh process, SAME key, SAME port, started immediately
$ MOSH_KEY=<KEY> mosh-client 192.168.0.180 60008
mosh: Nothing received from server on UDP port 60008.
```

Client B never receives a single byte. I ran it with a tight gap (server still bound, well inside any timeout) and got the same result, and Haven's Kotlin transport behaves identically against a live server — so it's the protocol, not the client implementation. The server's SSP/crypto state has moved past what a from-scratch client sends, so its datagrams are discarded as stale.

This is also why mosh upstream has no "reattach" command, and why the usual advice is to run tmux *inside* mosh.

### What that means for this PR

The persistence machinery itself works — I verified on-device that the saved (ip, port, key) survives an app restart, the re-attach path fires (`details: "mosh re-attach"` in the connection log), and my liveness probe correctly notices the server never answers, wipes the stale tuple and falls back to a normal SSH bootstrap. So it fails *safely*. But the user-visible result is: every post-restart connect stalls ~10s probing a server that can never answer, and then starts a fresh session anyway — strictly worse than today's instant fresh session, for zero benefit.

So I'm not going to merge this, and I don't think it can be rescued in this shape. I'm sorry to land on that after you put three rounds into it — the design work was sound, and the reasoning ("the server outlives the client, so reuse it") is exactly what I'd have assumed too. The thing that beat us is a protocol constraint neither of us checked early enough. That's on me as much as anyone: I should have tested two-clients-one-key before encouraging the direction in #365.

### What actually survives an app restart today

- **A session manager inside mosh** (tmux/zellij/screen — Haven supports this per-profile). The mosh session dies with the client, but `mosh-server` + `tmux attach` gets the *shell state* back, which is what users actually want. This already works.
- **Eternal Terminal**, which was designed for exactly this and *does* support reattaching a new client to a live session. Haven supports ET as a transport.

If you'd like to keep going on this problem, the promising direction is making the tmux-inside-mosh path automatic and obvious (e.g. default a mosh profile to a session manager, and on reconnect attach to the profile's `lastSessionName` without prompting) — that delivers the actual goal, "my session is where I left it", without needing mosh to do something it can't. Happy to support that if you want to take it on.

The finished implementation is on this branch (`4aeb1753`) if you want to read it, but I'd suggest we close this PR. Thanks again for the persistence on it — the #365 work that came out of this thread was a genuine win and is shipping.

GlassOnTin:
Closing this one — the approach can't work, for a reason neither of us saw coming.

A mosh-server will not accept a second client process, even with the correct key, port and IP. The session is bound to the first client; a fresh client's packets are ignored outright. That's stock mosh behaviour, nothing to do with Haven (repro in the comment above), and it means "the server outlives the app, so re-attach to it with the saved key" is not achievable. I should have tested two-clients-one-key back in the #365 round before encouraging this direction — that's on me, and it cost you three rounds of real work. Sorry.

What did come out of it:

- **#365 shipped** and is in users' hands (v5.68.39) — mosh sessions now survive network outages of any length. That was your fix, and it's the one people actually feel day to day.
- **The real "session where I left it" now works**, and it landed on main today as `ccc32cb4`: a Mosh/ET reconnect attaches straight back into the profile's remembered tmux/zellij session instead of stopping at the picker. The transport can't be resumed, but the multiplexer session behind it survives the app, so the shell comes back exactly where it was. Device-verified: marker typed → disconnect → reconnect → no picker, same tmux session, marker still on the pane. The picker still appears where you genuinely have a choice (nothing remembered, remembered session gone, or several remembered).

So the goal you were chasing is delivered — just through tmux rather than through mosh, because mosh won't have it.

Thanks for pushing on this. The debugging you did on the transport in #365 is still paying off, and I'd welcome your next one.

PR #412, opened 2026-07-17
fix(sftp): handle JSch IOException('inputstream is closed') with channel reset and retry

mintleaf84:
Fix for the `Failed to list: java.io.IOException: inputstream is closed` error when connecting to folders, particularly on local VM SSH servers.

**Changes:**

1. **JschSftpSession** — Add `IOException` catch to `translatingJschErrors()` so JSch's raw IOExceptions are wrapped in `SshIoException` instead of propagating through
2. **JschSftpSession** — Call `channel.setBulkSize(16KB)` in init to reduce fragmentation-triggered `fill()` failures on VM SSH servers (mwiede/jsch#858)
3. **SftpViewModel** — Catch `SshIoException` in `loadDirectoryEntries()` and `openSftpAndList()`, reset the corrupted SFTP channel, and retry once before surfacing error to the user

**Root cause:** JSch's `ChannelSftp.fill()` throws `IOException("inputstream is closed")` when `io_in.read()` returns `<= 0`, which can happen with fragmented SFTP responses, VM SSH servers, or Android socket behavior differences. The JSch library has a known issue (mwiede/jsch#858) with this behavior.

PR #416, opened 2026-07-19
fix(sftp): catch pre-warm SFTP session failure in selectProfile cached branch

mintleaf84:
Follow-up to #412.

**Problem:** The cached profile branch in `selectProfile()` launches a fire-and-forget coroutine to pre-warm the SFTP session, but JSch's `ChannelSftp` can throw `IOException("inputstream is closed")` during the INIT handshake (mwiede/jsch#858). Without a try-catch, this propagates to the uncaught exception handler and crashes the app.

**Fix:** Add try-catch matching the existing codebase pattern (same as the rclone capabilities pre-warm at the same call site). The failure is logged and silently swallowed — the next explicit operation (`openSftpAndList` / `loadDirectoryEntries`) will open a fresh session with proper error/retry handling.

**Root cause:** PR #412 fixed `openSftpAndList` and `loadDirectoryEntries` but missed the fire-and-forget pre-warm coroutine on line 1442.

### DIFF
--- a/feature/sftp/src/main/kotlin/sh/haven/feature/sftp/SftpViewModel.kt
+++ b/feature/sftp/src/main/kotlin/sh/haven/feature/sftp/SftpViewModel.kt
@@ -1440,7 +1440,11 @@ class SftpViewModel @Inject constructor(
                 isReticulum -> { /* backend resolved on demand via TransportSelector */ }
                 else -> {
                     viewModelScope.launch(Dispatchers.IO) {
-                        sftpSession = sessionManager.openSftpSession(profileId)
+                        try {
+                            sftpSession = sessionManager.openSftpSession(profileId)
+                        } catch (e: Exception) {
+                            Log.w(TAG, "Failed to pre-warm SFTP session for $profileId", e)
+                        }
                     }
                 }
             }


mintleaf84 (commented):
I've reviewed this PR. Here are my observations:

**What the review (4730041392) says:**
Copilot was unable to review due to quota limits — it's a non-review with no actionable feedback.

**My review of the PR itself:**

The change looks correct and follows the pattern described. A few observations:

1. **Matches existing pattern** — The try-catch +  approach is consistent with the rclone pre-warm at the same call site, as the PR description states.

2. **Correct semantics** — Catching  (broad) is appropriate here since JSch can throw various unchecked exceptions from the SFTP handshake. Silently swallowing is correct for a fire-and-forget pre-warm cache — the subsequent explicit operations have their own error handling.

3. **No resource leak** — If  throws during construction, no valid session reference is assigned, so there's nothing to clean up.

4. **Nit (minor):** Consider whether  could be left in an inconsistent state if a previous valid session existed. Looking at the code, this branch ( with cached profile) likely means this assignment overwrites any prior session. If the pre-warm fails, the old (possibly stale) session remains — which may or may not be desirable. However, this is consistent with the existing behavior before the fix.

**Verdict:** The fix is correct, minimal (+5/-1), and addresses the regression introduced by #412. Approve.

mintleaf84 (commented):
I've reviewed this PR (and the copilot review at #pullrequestreview-4730041392).

**What the copilot review says:** It's a placeholder — Copilot couldn't review due to quota limits, so no actionable feedback there.

**My review of the PR itself:**

The change looks correct and follows the described pattern:

1. **Matches existing pattern** — try-catch + `Log.w` is consistent with the rclone pre-warm at the same call site.
2. **Correct semantics** — Catching broad `Exception` is appropriate here; JSch can throw unchecked exceptions during SFTP INIT handshake. Silently swallowing is correct for a fire-and-forget pre-warm cache — subsequent explicit operations have their own error/retry handling.
3. **No resource leak** — If `openSftpSession` throws, no valid session reference is assigned, so nothing to clean up.
4. **Minor consistency note:** If a prior valid `sftpSession` existed, a failed pre-warm leaves the stale session in place. This is the same behavior as before the fix and is acceptable — the stale session will be replaced on the next explicit operation.

**Verdict:** The fix is correct, minimal (+5/-1), and properly addresses the regression from #412. I'd approve.

Issue #231, opened 2026-06-08
[feature request] Add custom labeling of the SSH keys


vintozver:
It's always good to have multiple hardware keys for the same identity, just for the case if one is lost.
Each ssh-keygen-produced and subsequently imported FIDO2 key has the following items:
* FIDO2: ssh:_{application}_
* sk-ssh-_{scheme}_@openssh.com
* SHA256:_{hash}_
* date when imported

If multiple keys are imported for the same identity, it's only possible to distinguish the key by the order they were imported, or a hash value.
Obviously, application is identical (key use domain), scheme is the same, and hash value is not something convenient to recognize.
A label would be great. It can be filled default from the comment section of the key, editable at the time of import, or with a long tap later.

For example, my setup is:
- [ ] YK vpn carry A (dongle 1)
- [ ] YK vpn carry C (dongle 2)
- [ ] YK vpn station A (dongle 3)
- [ ] YK vpn station C (dongle 4)
- [ ] YK pi carry A (dongle 1)
- [ ] YK pi carry C (dongle 2)
- [ ] YK pi station A (dongle 3)
- [ ] YK pi station C (dongle 4)

Thank you for considering the feature!

vintozver:
Thank you! Verified, works!

Issue #58, opened 2026-03-31
Possible SSH library replacement


kruton:
Hi, I see you're using termlib in Haven which is great. I'm also writing another SSH library in Kotlin which you might be interested in. It's currently released in Maven as [`org.connectbot.sshlib:sshlib`](https://central.sonatype.com/artifact/org.connectbot.sshlib/sshlib/versions) if you want to check it out. Currently the repo is at https://github.com/kruton/ssh-proto/ but will move to ConnectBot org soon.

kruton:
> A few questions:
> 
> * Would you welcome PRs for exec and SFTP support?

Yes, I think the SSH state machine itself is pretty much complete. Adding exec is just another channel on top of that.

> * Is there a roadmap or areas you'd prefer to tackle yourself vs. accept contributions?

Not really, but I would appreciate opening an issue to have an overview of the proposed architecture for a large feature before work starts.

> * Any architectural preferences for how SFTP should sit on top of the channel layer?

For SFTP you might need to create another state machine to drive the SFTP subsystem over the SSH channel. I haven't looked closely at how SFTP operates, but I recall that there are several versions and support for the different versions are widely varied.

I wouldn't mind supporting multiple different versions depending on the server being connected to.

> I was thinking of starting with the exec channel as a first PR to get familiar with the codebase, then tackling SFTP as the bigger effort. Haven would stay on JSch in the meantime and switch over once these features are stable.

That sounds reasonable! I have been testing ConnectBot using a locally compiled version with the new ssh-proto library to work out any of the bugs.

GlassOnTin:
Status update on the ssh-proto migration path:

| PR | State |
|---|---|
| ssh-proto#95 (exec channel) | ✅ MERGED |
| ssh-proto#98 (subsystem channel) | ✅ MERGED |
| ssh-proto#112 (SFTP client) | 🔄 review-addressed, awaiting maintainer re-review |

Just pushed a follow-up to #112 that addresses kruton's earlier
"don't throw from the suspend API" feedback — the `SftpPacketIO`
framing layer was the last throwing site; it now returns
`SftpResult<…>` like everything else on the suspend surface.
Throws are scoped to the `BlockingSshClient` Java-interop wrapper
where they belong. 426 unit tests pass locally.

Once #112 lands and a release goes out, the remaining path to
Haven adoption is:

1. FIDO2/SK auth extension on ssh-proto (separate PR; needed for
   Haven's hardware-key support).
[...]

PR #436, opened 2026-07-22
feat(ssh): add profile remote command (attach-or-create tmux via SSH exec, mosh-compatible)

kanazawahere:
## Summary
- Add per-profile `remoteCommand` and PTY preference (`requestPty`), including Room migration and encrypted backup round-trip.
- Open SSH `exec` channels for the configured command instead of an interactive shell — this bypasses login-shell startup files (`.bashrc` etc.), which previously raced with server-side tmux auto-wrap hooks when using `postLoginCommand`.
- Pass the profile command through to Mosh via `mosh-server ... -- <command>`, so mosh's network resilience (WiFi/cellular roaming) is preserved — unlike server-side forced-command SSH restrictions, which are incompatible with mosh's bootstrap.
- Adds a "Remote command (advanced)" field + "Request terminal (PTY)" toggle to the connection editor (hidden for ET transport, which has no equivalent bootstrap contract).

## Motivation
Typical use case: `tmux new -A -s <name>` as the remote command reliably attaches to (or creates) a specific named session every time a connection profile is opened — no risk of racing a shell startup hook into the wrong/duplicate session, and no risk of losing mosh.

## Verification
- Source gates: `scripts/check-i18n-hardcoded.sh`, `scripts/check-changelog.sh`, generated `docs/i18n/strings.json`, and `git diff --check`.
- CI (fork run): checks/test/supply-chain/build/lint/R8-smoke all green.
- Hands-on device test: verified end-to-end on a real Android device against a live server — connection profile correctly attached to the target tmux session, confirmed via a live round-trip message sent from the server side.


GlassOnTin:
Thanks for this — the design is sound (exec channel to skip the `.bashrc`/tmux-hook race, mosh `-- COMMAND` passthrough, clean v79→v80 migration, device-verified attach). I want it in, but it needs a rebase before it can land, because the SSH layer moved under it.

**What changed on main:** the #58 dual-SSH-engine work (JSch **and** a native sshlib engine, both behind a common `SshConnection`) landed after you branched. It deliberately **neutralised direct JSch usage** in `ShellChannel` / `SshClient` / `TerminalSession` — `ShellChannel` no longer references `com.jcraft.jsch.*` at all. Your commit adds `ChannelExec` / `ChannelSession` straight into `ShellChannel`, so those three files conflict head-on (the non-SSH parts — DB migration, editor UI, backup, deep-link `command=` param, mosh path — all merge cleanly; only the exec-channel plumbing collides).

**What the re-port needs** (smaller than it sounds — the extension points already exist):
- Add an exec-with-command path to the shell abstraction rather than opening a raw JSch `ChannelExec`. `SshConnection.openShellChannel(term, cols, rows)` is the seam — either extend it with an optional `command`/`requestPty`, or add a sibling `openExecChannel(command, requestPty, term, cols, rows)`.
- Implement it in **both** engines: JSch (`SshClient` — `openChannel("exec")` + `setCommand` + `setPty`, which you already had) **and** sshlib (`sshlib/SshlibShell.kt` — it already does `session.requestPty(term, cols, rows)`, so PTY is there; it needs the exec-command variant). If sshlib turns out not to support exec-with-PTY the way JSch does, gating the feature to the JSch engine is an acceptable v1 — your call, but flag it rather than silently no-op on sshlib.
- Extend the engine parity harness — `ShellChannelContractTest` / `ExecContractTest` in `core/ssh/src/test` — so the new path is covered for both engines. That contract suite is the gate that keeps the two engines honest; a green run there is most of the confidence, then your live-server tmux attach test on top.

If you rebase onto current `main` and push, I'll review and land it. If you'd rather I take the SSH-layer re-port, say so and I'll open a follow-up — but you own the live-server verification (I don't have your tmux setup), and the sshlib exec semantics are a judgement call I'd want your read on. Leaving this open for the rebase.

(Note: the floating-text-input PR #439 shared this branch's base, so I landed only its feature commit there and left the deep-link `command=` plumbing to this PR, where it belongs.)

GlassOnTin:
Rebased this onto current `main` and did the SSH-layer re-port myself (force-pushed to the branch). Everything outside the SSH shell layer — the DB `79→80` migration, the editor UI + PTY toggle, the encrypted-backup round-trip, the `haven://connect?command=` deep-link, and the mosh `-- COMMAND` path — merged cleanly and is **unchanged** from your commits, so your end-to-end device test of the attach UX still stands.

**What I changed vs your version** (only the exec channel, to fit main's #58 dual-engine refactor):
- Main neutralised `ShellChannel` — it's now streams + behaviour closures, no exposed JSch channel, and resize moved onto `ShellChannel`. So `openRemoteCommandOn` now builds the neutral channel (its `resizeFn` is `if (requestPty) channel.setPtySize(...)`), and the obsolete `resizeShell(ChannelShell)` is gone (main already calls `shell.resize(...)`).
- `openTerminalChannel(remoteCommand, requestPty, …)` is now a method on the `SshConnection` interface (implemented only by the JSch engine). The exec path is JSch-only for now: `SshEngine.SSHLIB` still builds a JSch connection (see `SshConnectionFactory`), and a sshlib `openTerminalChannel` is future #58 work — gated on sshlib maturity (the rekey fix in cbssh#231, exit-status in cbssh#232). Documented that on the interface.

**Verification**
- Compiles: `core:ssh`, `core:data`, `feature:connections`, `app`.
- Unit tests green: `core:ssh` full suite (engine contract tests + your `ShellChannelTest` case for `openRemoteCommandOn`, adapted to assert resize behaviour instead of the removed `.channel`), `core:data` (v80 migration + backup round-trip).
- Added **`JschRemoteCommandChannelTest`** — drives `openTerminalChannel` against a real MINA sshd whose `CommandFactory` echoes the command it received and whose `ShellFactory` writes a distinct banner, so the bytes prove the path: a command opens an `exec` request for exactly that command (incl. the `tmux new -A -s work` + PTY case), and null/blank falls back to an interactive shell.

Merging once the PR build goes green. Thanks for the feature — the exec-before-startup-files rationale is the right call.

kanazawahere:
Thanks for the rebase (38b123c9) — landed while I was independently building a fuller re-port overnight against the same #58 seam, so noting what's there in case it's useful, without touching your commit.

**On `feature/remote-command-profile-v2`** (this fork, not the PR branch — did not force-push over your fix): a version that additionally covers the sshlib engine and adds dedicated contract-test coverage.

**One finding worth flagging directly:** your commit's docstring says sshlib exec support is "future work... gated on sshlib maturity, cbssh#231". I checked the actual sshlib 0.3.1 sources (downloaded the sources jar and read `SshSession` directly) rather than assuming — `SshSession.requestPty(...)` + `SshSession.requestExec(command): Boolean` are both present and the pty-req + exec pairing works together on one session channel. I added `SshlibShell.requestExec(...)` (mirroring the existing `requestShell`) and a `RemoteCommandChannelContractTest` (same real-MINA-sshd pattern as `ShellChannelContractTest`, banner-distinguished so a wrong channel type fails on assertion rather than a hang) with both a JSch and an sshlib subclass — 385/385 `:core:ssh` tests pass locally, including both new subclasses. Exit-status still reports the documented `-1` placeholder on the sshlib side (cbssh#232, separate from #231), asserted rather than faked.

If that's useful, happy to open it as a follow-up PR once #436 lands, or point you at the branch directly — your call, not trying to override the minimal fix you already shipped. Full plan/reasoning is in `docs/plans/remote-command-ssh-engine-report.md` on that branch if it's useful context.

GlassOnTin:
Device-verified end-to-end in the shipped release (**v5.83.0**) on a OnePlus 13, against a live server (real SSH host, `192.168.0.180`):

1. **Install/migration** — installed the release-signed v5.83.0 `-r`; the app launched cleanly, so the `79→80` migration ran on the real on-device DB (Room would've crashed otherwise), no migration errors in logcat.
2. **Setup** — created an SSH profile, set **Remote command (advanced) = `tmux new -A -s attachtest`** in the editor (PTY defaults on), saved.
3. **Connect → exec path** — the session came up `sessionManager: NONE` (i.e. the exec channel, not the tmux-via-sessionManager path) and landed **directly in tmux**: the pane showed a bash prompt with the tmux status bar `[attachtest] 0:bash*  "…" …`, plus `mouseMode:1002` / `bracketPasteMode` that tmux sets — not a login shell.
4. **Attach-or-create / reconnect** — dropped a marker in the session, disconnected, reconnected: a fresh SSH session re-ran the exec command and **re-attached to the same `attachtest` session** (marker intact, no duplicate) — exercising the reconnect exec-channel path in `SshSessionManager`.

So the feature works on-device exactly as intended, on top of the `JschRemoteCommandChannelTest` MINA-server unit test and your own device test.

One follow-up came out of it: the remote command was only settable in the UI over the agent bridge, so `create_connection` / `update_connection` now take `remoteCommand` + `requestPty` and `list_connections` reports them (shipped in **v5.83.1**).

Thanks again for the feature.

PR #439, opened 2026-07-22
feat(terminal): floating text input dialog (ported from ConnectBot)

kanazawahere:
## Summary
- Ports ConnectBot's `FloatingTextInputDialog` (Apache-2.0) into Haven's terminal feature — a
  draggable/resizable floating dialog for typing longer text and sending it to the active
  session, useful when the on-screen keyboard makes multi-line composition awkward.
- Adds a new `TEXT_INPUT` keyboard-toolbar key (reorderable like any other toolbar key) that
  opens the dialog.
- Draft text is hoisted per-tab (keyed by `sessionId`), so switching tabs mid-draft doesn't lose
  or leak text into the wrong session.
- Send path reuses the existing bracket-paste-aware `sendInput`, so multi-line text pastes
  correctly into apps like vim/psql that honor bracket-paste mode.
- Move/resize are exposed to TalkBack via `semantics { customActions = ... }`.
- Upstream license header retained in the ported file per Apache-2.0 §4; see
  `docs/plans/floating-text-input.md` for the full design writeup and the red-team review this
  went through before implementation.

## Known CI status
- `build` (debug APK) passes.
- `lint` currently fails on `MissingTranslation` for the new strings — per
  `docs/features/i18n.md` this repo's convention is new strings ship English-only and pick up
  translations via the community translate page, not hand-authored in the PR
  (`docs/i18n/strings.json` entries are intentionally left with empty `"t": {}` and say so in a
  comment). Flagging this explicitly rather than leaving the failing check unexplained — let me
  know if this project actually wants placeholder translations included in-PR instead and I'll
  adjust.

## Test plan
- [x] Built via CI (`workflow_dispatch`), installed the resulting debug APK on a real device,
      verified: toolbar key appears after enabling it in Settings → Keyboard & input → Keyboard
      toolbar (pre-existing saved toolbar layouts don't auto-gain new keys — expected, not a
      bug), dialog opens, typed text sends correctly to a live SSH session.
- [ ] Bracket-paste behavior inside vim/psql on real device (multi-line send) — not yet
      exercised beyond a single-line smoke test.
- [ ] TalkBack move/resize custom actions — not yet verified with a screen reader.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ the corresponding GitHub Release; a release can't ship without its section
 (enforced by `scripts/check-changelog.sh` in CI). The GitHub "Full Changelog"
 compare link is appended automatically — don't add it here.
 
+## v5.82.0
+
+⌨️ **Terminal: floating text input** — a new toolbar key opens a draggable, resizable text box floating over the terminal, so you can compose a whole command with your normal keyboard (autocorrect, swipe typing, voice input, cursor movement) and send it in one shot, instead of fighting the raw terminal cell character-by-character. Embedded newlines and tabs show inline as ↩ / ⇥ so you can see exactly what will be sent, and the text is bracketed-paste-wrapped on send, so a multi-line block arrives as a paste instead of executing line-by-line. Unsent drafts are kept per tab and survive rotation; the window position/size is remembered. The key sits on the default toolbar and can be moved/hidden like any other key. Ported from ConnectBot's Text Input dialog (Apache-2.0).
+
 ## v5.81.6
 
 🖱️ **Terminal: touchpad two-finger scroll on some tablets** — on certain OEM tablets (reported on OPPO Pad 3 Pro) a Bluetooth-keyboard touchpad's two-finger scroll moved the app's own lists but did nothing in the terminal. Those touchpad drivers only send a scroll to a view that advertises itself as scrollable, which the terminal wasn't doing. The terminal now declares vertical scroll semantics (a mouse wheel already worked), which should route the gesture through — and, as a bonus, makes the terminal scrollable to TalkBack. Candidate fix, since I can't reproduce it here — feedback welcome. (#419, thanks wxjiee)
--- a/app/src/main/kotlin/sh/haven/app/ConnectDeepLink.kt
+++ b/app/src/main/kotlin/sh/haven/app/ConnectDeepLink.kt
@@ -25,7 +25,7 @@ object ConnectDeepLink {
         val transport: String?,
         val session: String?,
         /** Optional remote command (Tin attach / clean tmux). `command` or `startupCommand` query. */
-        val command: String?,
+        val command: String? = null,
     )
 
     /**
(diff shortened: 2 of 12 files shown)


GlassOnTin:
Landed on `main` as 29460c20 — thanks, this is nicely done.

**One deliberate change from the PR branch:** I cherry-picked only the floating-text-input commit, **not** the branch's base commit `feat(atp): keep Tin clean-attach on kho-chau sync…` (the `haven://connect?command=` deep-link + `mosh-server -- COMMAND` plumbing). That piece belongs to the remote-command work in #436, not to this feature — it rode along only because this branch was cut from that base. Keeping them separate so #436 lands as its own reviewed change. The floating-text-input commit is fully self-contained (terminal + toolbar + i18n + design doc only), so it cherry-picked clean onto current main.

**What I verified before landing:**
- Code review of `FloatingTextInputDialog.kt` (license header retained per Apache-2.0 §4; state properly hoisted; the ↩/⇥ `VisualTransformation` offset-mapping is correct) and the `TerminalScreen` wiring.
- The send path is byte-identical to clipboard paste: both now call the extracted `bracketPasteWrap(text, isBracketPaste)`. So the multi-line "arrives as a paste, not line-by-line" behaviour you flagged as only single-line-tested is the **same code path** already device-verified for clipboard paste into vim/psql — that retires most of that risk.
- `feature:terminal` + `core:toolbar` lint (green after the AGP-9 `MissingTranslation` fix that just landed on main — that's why lint was red on your run) and terminal unit tests pass.

**Still worth a real device pass (didn't block landing, low blast radius — an opt-in toolbar key):**
- TalkBack move/resize custom actions with an actual screen reader.
- A live multi-line send into vim/psql (byte-path is shared with paste, but exercising it end-to-end through the dialog is still worth doing).

The `## v5.82.0` CHANGELOG section ships with the next release. Closing as landed.

GlassOnTin:
Device-tested the multi-line path end-to-end on the OnePlus 13 (release-signed build of `29460c20`), against a local PRoot Alpine shell with `bracketPasteMode` on:

1. Added the Text input key to the toolbar → tapped it → floating dialog opened over the terminal. ✓
2. Typed `echo AAA` ⏎ `echo BBB` — the embedded newline rendered as the inline `↩` symbol exactly as designed. ✓
3. Tapped Send. The terminal received both lines as **one bracketed paste** — they sat at the prompt unexecuted (no `AAA` output on Send); if the wrap were broken, `echo AAA` would have run immediately. ✓
4. Pressed Enter → both executed, output `AAA` then `BBB`. So the multi-line text was delivered intact and the paste-vs-line-execute behaviour is correct. ✓

That closes the "multi-line send not yet exercised beyond single-line" gap from the PR description.

Two minor observations (not blockers, polish for a follow-up):
- The dialog does **not** dismiss on Send — it clears the draft and stays open (fine if intentional for sending several lines; worth a deliberate decision).
- The dialog is a `Popup` that isn't torn down when you leave the Terminal tab, so if you open it and then switch to e.g. Settings, it keeps floating over that screen. Tearing it down on tab-change (or dismissing on Send) would avoid that.

TalkBack move/resize custom actions still unverified with a real screen reader — the only remaining item from the test plan.

PR #444, opened 2026-07-23
fix(terminal): text-selection toolbar missing in floating text input

kanazawahere:
## Summary

The floating Text Input dialog (`FloatingTextInputDialog.kt`, landed in #439) lets you select
text — selection handles appear — but the native Copy/Cut/Paste/Select-all toolbar never shows
up, found on a real device.

## Root cause

The dialog's `TextField` lives inside a focusable `Popup` (deliberate — see #439's design doc,
isolated pointer-input dispatch). Disassembled the actual pinned `androidx.compose.foundation`
1.11.4 AAR (via compose-bom 2026.06.01) to confirm rather than guess:
`ComposeFoundationFlags.isNewContextMenuEnabled` compiles to `true` by default, and when true the
selection machinery never calls `TextToolbar.showMenu()` — it routes through a parallel
`TextContextMenuProvider` pipeline instead. The platform default provider for *that* system
(`AndroidTextContextMenuToolbarProvider`) is `ActionMode`-based, which silently no-ops inside a
focusable `Popup` (no real `DecorView`) — same failure class as the legacy path, just on the
system that's actually active on this Compose version.

Material3's `TextField` follows the same path: M3 1.4.0 has no context-menu wiring of its own and
delegates to the legacy `CoreTextField`, which is flag-gated the same way.

## Fix

`FloatingInputTextContextMenuProvider` (a `TextContextMenuProvider`) renders the menu as a sibling
inside the dialog's own popup window instead of via `ActionMode` — no second focusable window, so
no recursion of the original problem. Menu items (labels, applicability, click behavior) come
straight from the field itself via `TextContextMenuDataProvider.data()`, so it automatically tracks
whatever the platform TextField would normally offer (Cut/Copy/Paste/Select-all, plus e.g. "Read
aloud" where available) rather than a hardcoded 4-item list.

Provided for **both** `LocalTextContextMenuToolbarProvider` and `LocalTextContextMenuDropdownProvider`
— the field's internal default-installer only respects an outside provider when both locals are
non-null (verified in the 1.11.4 bytecode), so providing only one gets silently shadowed back to
the broken platform default.

Kept a `LocalTextToolbar` implementation too (small, already tested) as a fallback for Compose
configs where the flag is false — dormant on this pinned version, cheap to keep.

No `ComposeFoundationFlags` mutation anywhere — the whole fix is scoped to composition locals
around the one `TextField`, nothing else in the app is affected.

## History (kept honest rather than squashed)

First attempt (this PR's first commit) shipped only the `LocalTextToolbar` override, reasoned to
be correct, unit-tested, and confirmed on a real device to do **nothing** — the flag findings
above are exactly why. Second commit is the actual fix, verified on a real device this time
(screenshot: Cut/Copy/Paste/Select all/Read aloud all appear and are positioned correctly over the
selection).

## Test plan

- [x] Real device: select text in the dialog → toolbar appears positioned over the selection,
      each button works (screenshot confirms Cut/Copy/Paste/Select all/Read aloud all present).
- [x] `feature:terminal` unit tests: 104/104 pass (10 new state-contract tests across both
      providers).
- [x] Confirmed unchanged: bracket-paste send path, drag-to-move, resize, TalkBack custom actions,
      tap-outside-dismiss.
- [ ] TalkBack with the new menu specifically (pre-existing gap from #439, not newly introduced
      here) — not yet verified with a real screen reader.

### DIFF
--- a/docs/plans/floating-text-input-toolbar-fix.md
+++ b/docs/plans/floating-text-input-toolbar-fix.md
@@ -1,6 +1,63 @@
 # Plan — custom text-selection toolbar for the floating text input dialog
 
-## Context
+## ⚠️ ROUND 2 (2026-07-23) — round-1 fix FAILED real-device test, new lead found
+
+The round-1 implementation (custom `TextToolbar` via `LocalTextToolbar`, committed as
+`f420edf1` on `fix/floating-text-input-selection-toolbar`) compiled clean, passed its own unit
+tests, and was reasoned to be correct — but **Batin tested the real debug APK on his device and
+the toolbar still does not appear.** Selection still works (handles visible), still no
+Copy/Cut/Paste/Select-all menu. This is a genuine implementation failure, not a test error — do
+not re-ship round 1 as-is.
+
+**New lead, found by reading real Compose Foundation source** (`androidx/androidx` on GitHub,
+`compose/foundation/foundation/.../text/selection/TextFieldSelectionManager.kt` and
+`ComposeFoundationFlags.kt`): recent Compose Foundation has **two parallel context-menu systems**
+gated by `ComposeFoundationFlags.isNewContextMenuEnabled` (default comes from a per-platform
+`internal expect val isNewContextMenuInitiallyEnabled`, not confirmed for Android/this exact
+version via web source alone — **verify the REAL compiled default by decompiling the actual
+pinned `androidx.compose.foundation` AAR classes in this machine's Gradle cache, not by trusting
+a web search** of the source, which hit rate limits and inconclusive results this round).
+
+The doc comment on the flag: *"Whether to use the new context menu API and default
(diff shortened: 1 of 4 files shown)


GlassOnTin:
Merged and shipped in v5.83.4 — thanks @kanazawahere.

Verification: both new unit suites pass (5 FloatingInputTextContextMenuProviderTest + 4 FloatingInputTextToolbarTest, 9/9), on top of the bytecode-level root-cause analysis in the PR.

Full disclosure on device testing: I couldn't capture an on-device screenshot of the selection toolbar inside the floating input dialog. The dialog is opened by a keyboard-toolbar key, and our synthetic-tap MCP verb can't actuate the custom-drawn toolbar keys (a separate injection limitation, now flagged for follow-up), so I couldn't drive it remotely. The fix therefore rides on the unit tests + the root-cause analysis rather than a device capture.

PR #447, opened 2026-07-24
feat(terminal): Save connection pins live multiplexer session to a profile

kanazawahere:
## Summary

Adds **Save connection** from a live terminal tab so a multiplexer session can be pinned as a one-tap Connections profile (same idea as hand-edited `remoteCommand` role profiles).

### UX
1. Long-press a terminal tab
2. **Save connection**
3. Dialog: **Name** (editable, default = session name) · **ID** (read-only profile id) · **OK** / **Cancel**
4. OK → upsert profile in Connections (clone host/auth/mosh from live session + pin)

### Behaviour
- Pin via `remoteCommand` (`tmux new -A -s …` / zellij / screen / byobu) + `lastSessionName` + `requestPty`
- Upsert key: same **label + host + username + port** → update pin; else create new UUID
- Distinct from `lastSessionName` reconnect memory (auto remember); this is an explicit save that closes the loop
- Works for SSH session-manager attach and for profiles already using `remoteCommand` / Mosh / ET

### Tests
- Unit tests: `SaveConnectionFromSessionTest` (pin strings, draft/upsert, parse remoteCommand)

## Test plan
- [ ] Connect with session manager (tmux), attach a named session
- [ ] Long-press tab → Save connection → edit name → OK
- [ ] Connections list shows the new card; cold connect auto-attaches without picker
- [ ] Save again with same name updates pin (no duplicate)
- [ ] Cancel leaves DB unchanged
- [ ] CI unit tests green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ the corresponding GitHub Release; a release can't ship without its section
 (enforced by `scripts/check-changelog.sh` in CI). The GitHub "Full Changelog"
 compare link is appended automatically — don't add it here.
 
+## Unreleased
+
+🔌 **Terminal: Save connection pins the live session as a one-tap profile** — long-press a terminal tab → **Save connection** opens a confirm dialog (editable **Name**, read-only **ID**, OK/Cancel). OK clones the live SSH/Mosh/ET host/auth into a Connections profile and pins the multiplexer session via `remoteCommand` (`tmux new -A -s …` / zellij / screen / byobu), so a cold tap re-attaches without the session picker. Same name + host + user upserts the pin instead of duplicating. Distinct from `lastSessionName` reconnect memory: this is an explicit save that closes the loop. (thanks kanazawahere)
+
 ## v5.83.10
 
 🌐 **Local Linux: the guest DNS setting now also covers package installs** — v5.83.9 made the guest's name servers configurable, but only refreshed `/etc/resolv.conf` when a *terminal* session started or a distro was installed. Installing a desktop (or anything else run inside the guest) goes down a different path, so on an existing distro those commands could still use the stale file — which is exactly the "installing Xfce silently hangs" case the setting was added to fix. Every command run in the guest now refreshes the resolver first. (#446)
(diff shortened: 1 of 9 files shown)


kanazawahere:
### Follow-up fix (empty shell / clobbered fleet pin)

Dogfood found that **Save connection** rewrote custom `remoteCommand` wrappers (e.g. `bash …/haven-role-attach dogfood` for conversation recovery) back to plain `tmux new -A -s …`, which left cold taps on an empty shell when the agent process was not already live.

This push:
- Preserves non-standard / fleet remoteCommand on upsert
- Still uses simple multiplexer pins for new standard cards
- Unit tests for wrapper preservation

CI will re-run on this branch.

GlassOnTin:
Merged — this is on `main` as 5fd78349, with you as the commit author.

Thanks for this. Pinning a live multiplexer session as a one-tap profile closes a real gap, and the upsert on label+host+username+port is the right call: it keeps repeat saves from breeding duplicates, which is exactly where this kind of feature usually goes wrong.

## About the red CI

Worth saying explicitly, because it was not your change: the failing job was `SshlibExecContractTest > stderr and non-zero exit status`, a flake that had been hitting main intermittently. I root-caused it the day after your run — sshlib hands command output over a rendezvous channel and cancels the pump on channel close, so a drain that had not yet parked lost the output entirely. Fixed on main, and reported upstream as connectbot/cbssh#245. Your branch simply had the bad luck of being tested before that landed.

## What I changed while rebasing

The branch had diverged over a run of releases, so I applied it to main rather than merging, and dropped two commits:

- **`@SuppressLint("MissingPermission")` on ProotDns** — no longer needed. `core/local` now declares `ACCESS_NETWORK_STATE` in its own manifest, so lint is satisfied when it analyses the library in isolation without suppressing the warning. Your instinct was right that lint needed appeasing; the manifest is the more durable place to do it.
- **The CHANGELOG entry** — rewritten for whichever release this lands in, since CHANGELOG conflicted heavily.

`docs/i18n/strings.json` is regenerated from your new strings rather than carried across, for the same reason.

Everything else is yours as written. Tests pass, i18n export is clean.

Closing this PR since the work is already on main.

Issue #359, opened 2026-07-09
[FR] Automatically synchronizes with existing connections


hung319:
Regarding the issue at https://github.com/GlassHaven/Haven/issues/323, Haven can now push or pull backups from remotes like rclone, ftp, and sftp. However, to make it easier, I suggest adding a mechanism to automatically push backups whenever changes occur or at regular intervals.

In my opinion, since Haven data can be very small, it's perfectly feasible to push backups after each change or use a waiting mechanism to consolidate all changes at once.

However, this is just my opinion, as someone who isn't directly involved in editing the project. So I think if you have a better idea to optimize this feature, feel free to use it. 

Thank You !

GlassOnTin:
Thanks for opening this as the follow-up from #323 — accepted, this is the layer I said would sit cleanly on the manual Push/Pull.

The shape I have in mind:

- **Auto-push on change, debounced.** You're right that the backup is small, so pushing after each change is feasible; a short debounce consolidates bursts (e.g. editing several connections in a row) into one write. A periodic fallback (WorkManager) can cover anything a change-trigger misses.
- **Push only.** Automatic *pull* is where the sharp edges live — last-write-wins silently overwriting local changes made while offline. Pull stays manual until there's a story for conflicts.

One design point to be transparent about: manual Push prompts for your backup passphrase every time, and the passphrase is what encrypts the blob. Automatic push can't prompt, so enabling it means Haven has to keep the passphrase (encrypted in the Android Keystore) for the background job to use. That's a deliberate, opt-in trade — the remote still only ever sees ciphertext, but the phone holds the unlock material rather than your head. I'll make that explicit in the setting rather than bury it.

Leaving this open to track. No timeline promise, but the plumbing (BackupSyncManager) was built with this in mind, so it's mostly scheduling + the passphrase consent UX.

GlassOnTin:
Implemented on main (`99e23ba5`), in the shape discussed above:

- **Auto-push, debounced.** Turn on **Push automatically** in Settings → Backup → Sync to a remote. A background job re-pushes the encrypted backup ~2 minutes after config changes (edits keep resetting the timer, so a burst of changes collapses into one push — your "waiting mechanism to consolidate"), plus a daily catch-up push for anything the change triggers miss.
- **Push only.** Pull stays a manual, eyes-open action, for the conflict reasons above.
[...]

PR #458, opened 2026-07-28
feat(backup): opt-in automatic pull + remember-password + manual-pull shortcut

kanazawahere:
## Summary

Backup sync currently supports automatic **push** (#359) but pull stays manual by design — the original rationale (see `BackupAutoSync.kt`) is that an unattended pull could silently overwrite local edits on a last-write-wins remote. That tradeoff doesn't fit every setup: if a device is a pure *consumer* of a backup authored elsewhere (a headless "source of truth" host), automatic pull is safe and saves a manual step every time the source changes.

This PR adds automatic pull as a **separate, off-by-default opt-in** alongside the existing push toggle — it doesn't touch push's behavior or its warning copy (well, it does — see below).

- **"Pull automatically" toggle** in the backup sync dialog, next to the existing "Push automatically" toggle. Off by default. Enabling it prompts for the backup passphrase (stored encrypted on-device, same mechanism as auto-push) and schedules a periodic `WorkManager` pull.
- **Configurable interval** — 15 minutes / 1 hour / 6 hours / 24 hours (15m is WorkManager's platform floor; the periodic worker reschedules cleanly when the interval changes).
- **"Remember this password" checkbox** on the manual Restore Backup dialog, reusing the same encrypted-passphrase field auto-pull already uses (no second storage location). Unchecked by default; when checked, the password field pre-fills on the next manual restore without re-typing — restore still requires tapping Restore, keeping it an eyes-open action.
- **Static Android App Shortcut** ("Pull Backup Now", long-press the launcher icon) — runs a one-off pull in the background using the already-remembered passphrase, reporting success/failure via Toast + notification. If no passphrase is saved yet, it opens the app straight to the restore-password dialog instead of failing silently.
- Fixed a stale line in the push-automatically hint ("Pull always stays manual") that this PR makes inaccurate — it now points at the new toggle instead.

## Two bugs found via on-device testing (not just unit tests)

1. The shared password dialog's `isExport` flag didn't exclude the new `EnableAutoPull` action, so turning on "Pull automatically" incorrectly demanded a *new* password + confirmation instead of the *existing* backup passphrase — the toggle was unusable as shipped in an earlier revision of this PR. Fixed by adding `EnableAutoPull` to the same exclusion as `Restore`/`PullRemote`.
2. The App Shortcut activity crashed on tap (`IllegalStateException: You need to use a Theme.AppCompat theme...`) — it extended `AppCompatActivity` but the manifest used the plain platform theme `Theme.Translucent.NoTitleBar` (matching `RestartActivity`, which is a plain `Activity` and doesn't need AppCompat). Switched the base class to `ComponentActivity`, keeping `lifecycleScope` without the AppCompat theme requirement.

Both were caught live on a physical device (not just `assembleDebug`/unit tests), then fixed and re-verified on-device.

## Testing

- `:app:testX64DebugUnitTest` (`BackupAutoPullTest`, including a new test for interval-change rescheduling) — passing.
- `:app:assembleArm64Debug` — clean build, CI green (build/test/lint/checks) after the fixes above.
- Manual on-device verification: enabled auto-pull with the fixed dialog (single password field + remember checkbox, matching Restore's UX), selected each interval option, triggered the App Shortcut with and without a saved passphrase (correct fallback to the password dialog when unset), confirmed the worker enqueues and reports success/failure via Toast without crashing.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -77,7 +77,17 @@
                     android:scheme="haven"
                     android:host="connect" />
             </intent-filter>
+            <meta-data
+                android:name="android.app.shortcuts"
+                android:resource="@xml/shortcuts" />
         </activity>
+
+        <activity
+            android:name=".backup.PullBackupShortcutActivity"
+            android:exported="true"
+            android:excludeFromRecents="true"
+            android:taskAffinity=""
+            android:theme="@android:style/Theme.Translucent.NoTitleBar" />
 
         <!-- One-shot resurrector for the restart_app MCP verb. Runs in its own
              process (:restart) so it can kill the main process — applying a
--- a/app/src/main/kotlin/sh/haven/app/HavenApp.kt
+++ b/app/src/main/kotlin/sh/haven/app/HavenApp.kt
@@ -45,6 +45,7 @@ class HavenApp : Application(), Configuration.Provider {
     @Inject lateinit var sessionManagerRegistry: sh.haven.core.ssh.SessionManagerRegistry
     @Inject lateinit var mailWatchManager: sh.haven.app.agent.mailrules.MailWatchManager
     @Inject lateinit var backupAutoSyncScheduler: sh.haven.app.backup.BackupAutoSyncScheduler
+    @Inject lateinit var backupAutoPullScheduler: sh.haven.app.backup.BackupAutoPullScheduler
     @Inject lateinit var sshTerminalEmulatorOwner: sh.haven.feature.terminal.SshTerminalEmulatorOwner
 
     private val appScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
@@ -246,6 +247,7 @@ class HavenApp : Application(), Configuration.Provider {
         // Auto-push backup watch (#359) — same pattern: inert until the user
         // enables auto-sync in Settings → Backup → Sync to a remote.
         backupAutoSyncScheduler.start(appScope)
+        backupAutoPullScheduler.start(appScope)
 
         // Extend the shell-prompt terminator set used for command-on-attach
         // detection with the user's custom prompt characters (#280). Replays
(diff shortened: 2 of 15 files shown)


GlassOnTin:
Reviewed and merging. This is careful work, and a couple of the details are ones I would expect a reviewer to have to point out rather than find already handled.

The one I want to name specifically: the shared passphrase key is only removed when **both** auto-push and auto-pull are off, and you added the reciprocal check to each setter. That is exactly the bug I would have gone looking for — turning one toggle off silently breaking the other — and it is already right. Likewise the overwrite risk is stated plainly in the hint text rather than buried, `ExistingPeriodicWorkPolicy.UPDATE` handles the interval change properly, and unticking "Remember this password" actually clears the stored value, with the checkbox reflecting whether one is stored. A remembered credential you cannot see or remove would have been a blocker; this is neither.

The write-up of the two on-device bugs is worth more than the fixes. "The toggle was unusable as shipped in an earlier revision of this PR" is the kind of thing most PRs quietly drop from the description, and both were things `assembleDebug` and unit tests could not have caught.

## Two things I am fixing myself rather than sending back

**The eleven new strings are English-only.** Haven reached complete coverage across all eleven languages yesterday, and `"t": {}` on each new key gives that back. Not your job to know that landed hours before you opened this — I will fill them.

**The worker's user-facing text is hardcoded.** `"Backup Sync Success"`, `"Restored N items"`, `"Backup restored successfully!"`, `"Backup pull failed: …"` and the `"Backup Sync"` channel name are English string literals in `BackupAutoPullWorker`. Worth flagging why this passed review-by-CI: `check-i18n-hardcoded.sh` only inspects Compose `Text(...)` and `title=`/`subtitle=`, so notification and Toast text walks straight past it. The existing push worker has no user-facing text at all, so this is new surface rather than an existing pattern you were matching. I will move these to resources and translate them in the same pass.

One trivial note, not worth a change on its own: the notification id `424242` is a magic constant that could collide with another notification later.

Merging as-is and doing the i18n follow-up on main. Thanks — this is the second one of these you have got right, and the on-device discipline is the reason it is mergeable.

PR #464, opened 2026-07-30
fix(backup): don't crash when the stored auto-pull passphrase can't be decrypted

kanazawahere:
## Summary

Found this live on a real device while testing #458 (auto-pull) end to end: the app crash-loops on every launch with `java.security.GeneralSecurityException: decryption failed` from `CredentialEncryption.decrypt()`, thrown uncaught inside `UserPreferencesRepository.backupSyncPassphraseFlow` / `backupSyncPassphrase()`, which `BackupAutoPullScheduler.start()` collects from `HavenApp.onCreate()`.

**Root cause:** the stored passphrase ciphertext is Android Keystore-backed, and Keystore keys are deliberately excluded from Android's own app-data backup/restore mechanism. An app reinstall or device data restore can leave the old encrypted DataStore value behind while the Keystore key itself doesn't survive — so it can never be decrypted again. This isn't specific to any one backup file; it's a real scenario any user can hit (reinstall, new device, etc.), and once it happens the app becomes unusable until the user clears app data.

**Fix:** treat an undecryptable stored passphrase the same as "none configured" instead of letting the exception propagate. Wrapped in a small private helper so both the `Flow` and the suspend accessor share the same handling.

## Test plan
- [x] Reproduced the crash on a real device with a stored (undecryptable) passphrase
- [x] Applied the fix, confirmed the app no longer crashes on launch with the same device state
- [x] CI green on this branch: build, test, lint, checks, nightly-abi (x64 + armv7), release-r8-smoke, supply-chain all pass

Note: an unrelated flaky test (`SshlibExecContractTest`, module `core:ssh`, a real-SSH-server integration test) is not touched by this change.

### DIFF
--- a/core/data/src/main/kotlin/sh/haven/core/data/preferences/UserPreferencesRepository.kt
+++ b/core/data/src/main/kotlin/sh/haven/core/data/preferences/UserPreferencesRepository.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import sh.haven.core.security.CredentialEncryption
+import java.security.GeneralSecurityException
 import javax.inject.Inject
 import javax.inject.Singleton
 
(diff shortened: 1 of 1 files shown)


GlassOnTin:
Merged — thank you, and especially for the root-cause paragraph. I verified it rather than taking it on trust: the reported `GeneralSecurityException: decryption failed` is Tink's AEAD failing against a fresh Keystore key after restore, exactly as you described, and every consumer of the two accessors already treats null as "not configured" (both workers bail with `Result.success()`, the shortcut activity checks `isNullOrEmpty`, the settings StateFlow tolerates it) — so the null-fallback is safe end to end.

One follow-up landed on main right after the merge (`82e0fc1d`, I couldn't push to your branch — the fork rejected the maintainer push): the catch is widened from `GeneralSecurityException` to `Exception`. `CredentialEncryption.decrypt` has two more failure surfaces that produce the *same* launch crash-loop from the same restore scenario — `Base64.decode` throws `IllegalArgumentException` on a corrupt stored value, and Tink's `AndroidKeysetManager.build()` can throw `IOException` on an unparseable restored keyset. All of them mean "the stored value is unusable", which is the same reasoning `isEncrypted` in that file already applies with its broad catch.

This will be in the next release.

GlassOnTin:
Out in v5.86.3 — your fix plus the widened catch. Thanks again for the on-device find and the diagnosis that held up under verification.

kanazawahere:
Thanks for merging this, and for double-checking the root cause yourself instead of taking the writeup on faith. We pulled your follow-up (82e0fc1d, widening the catch from GeneralSecurityException to Exception) into our downstream branch so we stay in sync. I also just sent a collaborator invite with write access, so next time you can push straight to our PR branch instead of routing through your own main like this round.

PR #472, opened 2026-07-30
Draft: Card/Grid/List browse modes + optional tin-companion live preview

kanazawahere:
This is a draft PR, not a request to merge as-is. It adds new browse modes to the Connections tab, including a live terminal preview backed by an optional host-side daemon, and I want design feedback before we talk about landing any of it.

### What's in this PR

- Three new Connections-list browse modes, switchable from the top bar: Card (full-screen cards, swipe up/down between them), Grid, and List.
- Card and Grid show a live terminal preview snippet on each card: recent tmux pane output, auto-refreshed on a poll interval.
- The live preview needs an optional companion daemon running on the connection's host. No daemon on the host → the card just falls back to a static representation. Nothing breaks if you don't install it.
- The daemon is a zero-dependency Python script (stdlib only). It exposes a read-only-except-kill HTTP API (`GET /status`, `GET /shells`, `DELETE /shells/<name>`) behind a bearer token.
- Security model: fail-closed deny by default, one token per install, no CORS, and deliberately no arbitrary-command-execution route.
- The daemon isn't bundled in this PR: it's a separate public repo (https://github.com/kanazawahere/tin-companion), installed with a curl|bash one-liner or manually.
- Client-side, it follows the same pattern Haven already uses for mosh-server: detect the missing helper by its failure signature, show install instructions, degrade gracefully if it's absent.
- Branch: feature/tin-companion-generic on https://github.com/DigitalVersion/Haven. CI is green, and the new client/viewmodel code has its own tests (TinPreviewCompanionTest, TinPreviewFilterTest).

This is a bigger, more opinionated change than our earlier fixes. It reshapes the Connections tab and pulls in a new kind of optional external dependency. I'm not assuming any of this should land as-is. Consider it working code to argue with: push back on the scope, the security model, or whether a companion-daemon concept belongs in Haven at all.


GlassOnTin:
You asked for pushback, so here it is properly — this is a serious piece of work and it deserves argument rather than a polite nod.

## The main one: I don't think you need the daemon

Haven already has `execCommand(command, timeoutMs)` on `SshClient` (and `SshlibExec.exec` on the other engine). So the preview you're fetching over HTTP is also reachable as:

```
tmux list-sessions -F '#{session_name} #{session_attached}'
tmux capture-pane -p -t <target>
```

over the SSH connection the user has already configured and already trusts. That gets you the same pane text with:

- no install step on the host
- no bearer token to store, rotate or leak
- no listening port
- no second repository to keep alive
- and it works on every host the user can already reach, including ones behind a jump host, because it rides the connection Haven already knows how to build

The daemon buys a cheaper poll and a structured JSON shape. I don't think that's worth a new trust boundary on every server someone owns. If there's a case `capture-pane` genuinely can't serve — I can imagine one around cost when previewing many hosts at once — that's the argument to make, and it would be the thing that justifies the daemon rather than the convenience of the API shape.

## Security, concretely

`TinPreviewClient` opens an `HttpURLConnection` to a configured `baseUrl` and sends `Authorization: Bearer $token`. Nothing in the client requires TLS or an SSH tunnel. If `baseUrl` is ever `http://`, the token crosses the network in cleartext — and that token authorises `DELETE /shells/<name>`, so it isn't read-only in effect. "Read-only except kill" is still a remote kill primitive.

If the intent is that this only ever runs over a Haven local port-forward, that constraint has to be enforced in the client, not documented — Haven has `-L` forwarding already, so binding the daemon to loopback on the host and reaching it only through a forward would close this properly. As written the security model is a description rather than something the code holds you to.

And a `curl | bash` install is a large ask for a preview feature. Whatever we do here, I'd want the install to be something a person can read first.

## Scope

The diff touches 36 files, including `feature/sftp`, `backup/BackupAutoPull`, `SshKeyResolver`, `SilentSshDialer`, `HavenNavHost`, `shortcuts.xml` and a Room schema `80.json`. Very little of that is browse modes or previews — it looks like the same fork-main pollution that sank #481 and #483.

That matters more than tidiness here: I can't review a security-sensitive feature when a database schema bump is riding along in the same diff. **Never bump the Room schema in a feature branch** — it's a one-way door on any device that installs it, since Room refuses to open a database newer than the app.

## What I'd actually like

Split it:

1. **Card / Grid / List browse modes, no preview.** That's a self-contained UI change and genuinely interesting on its own. Worth checking against one existing constraint: on small portrait phones we deliberately keep to one pane at a time and avoid persistent chrome, so full-screen swipe cards need to earn their space against the list they replace.
2. **Live preview via `execCommand` + `capture-pane`**, as a separate change with its own argument about polling cost.
3. **The daemon**, only if (2) demonstrably can't work.

The first could land soon. The third I'm sceptical of, but I'd rather be shown wrong with a measurement than assume.

## Small thing

Several comments in the new files are in Vietnamese (`Ném exception khi lỗi mạng/HTTP/parse — caller xử`). The codebase is English throughout; worth converting so the next person maintaining it isn't guessing.

None of this is "no". You've built something real and asked the right question about whether it belongs — the honest answer is that the browse modes probably do, and the daemon needs to beat an option you already have in the codebase.

GlassOnTin:
Two corrections to my review — I got one thing wrong and overstated another.

**The Room schema. I was wrong, and it was the most alarming thing I said.** I told you never to bump the schema in a feature branch. You didn't bump anything: the diff *adds* `80.json` as a new file and doesn't touch `version = 80` in `HavenDatabase`. The database has been at 80 since `9ed9d722` (22 July); the exported JSON simply hadn't been committed yet at your branch point. Main has since committed the identical file in `65595cd5` (1 August). So that file is correct, and on a current rebase it will vanish from your diff entirely. Sorry — that warning would have had you strip something legitimate.

**The scope point was too broad.** Re-reading the new-file list, it splits in two:

- `backup/PullBackupShortcutActivity.kt`, `backup/TransportBackupIo.kt`, `BackupAutoPullTest.kt`, `shortcuts.xml` — unrelated to browse modes, and this is the part I'd want out of the diff.
- `SilentSshDialer.kt`, `SshKeyResolver.kt` and its impl/module — I lumped these in with the above, and shouldn't have. They read as load-bearing for the feature.

That second group actually strengthens the argument I was making. If you already have a silent dialler, then "connect quietly, run one command, drop" is a mechanism you've built — which is exactly what `execCommand("tmux capture-pane -p -t …")` needs. The daemon's advantage over that shrinks to poll cost, which is measurable rather than architectural.

Everything else in my review stands, particularly the bearer token over a plain `HttpURLConnection`. That's still the thing I'd want closed before any of the preview half lands, and Haven's existing `-L` forwarding is the tool for it: bind the daemon to loopback on the host, reach it only through a forward, and the token never crosses a network at all.

Issue #448, opened 2026-07-26
Flaky: SshlibExecContractTest loses one of stdout/stderr intermittently in CI (real stream loss, not test noise)


GlassOnTin:
## Symptom

`SshlibExecContractTest > stderr and non-zero exit status` fails intermittently in CI, failing the whole `test` job — and therefore any release build it lands in.

From the CI test report (run 30174838095, commit 17fdc86f):

```
org.junit.ComparisonFailure: expected:<[oops
]> but was:<[]>
	at sh.haven.core.ssh.ExecContractTest.stderr and non-zero exit status(ExecContractTest.kt:70)
```

**stderr came back empty.** The stdout assertion on the line above passed, and the exit-status assertion below was never reached. So this is not "the test is wrong" — a remote command's stderr was genuinely lost.

Seen at least twice: on this main run, and on the CI for #447 (unrelated changes, so it is not caused by either).

## What is ruled out

- **Not the exit-status path.** The failure is a String comparison; `exitStatus` is asserted afterwards and never got there.
- **Not cbssh#238 (second session channel).** Each test method builds a fresh client and runs exactly one `exec`, so only one session channel exists.
- **Not `SshlibExec`'s drain logic**, as far as I can see: stdout and stderr are drained concurrently, both are awaited, and the exit status is only read afterwards.
- **Not reproducible locally.** ~130 runs on an idle machine and under deliberate CPU contention (load average >25 on 12 cores) produced 0 failures. It appears to need the specific scheduling of a loaded CI runner.

## Likely mechanism (not proven)

In sshlib 0.4.0, `SessionChannel` does not hand the consumer the receive channel the packet loop writes to. It keeps `stderrIngress` and `_stderr` as separate `Channel`s with a `stderrDeliveryJob` coroutine pumping one into the other (same for stdout). If the channel's close path can run before that pump has drained `stderrIngress`, buffered stderr is discarded and the consumer sees a cleanly-closed, empty channel — which is exactly what the failure looks like. Under load the pump is the thing most likely to be descheduled.

That is inference from the class structure, not something I have demonstrated, and the fix (if it is this) would be upstream in connectbot/cbssh rather than here.

## Impact

Low for users right now — this is the **experimental** sshlib engine (`HavenSshEngine sshlib`); JSch remains the default and is unaffected. But if the mechanism above is right, trailing stderr can be silently lost from remote commands on that engine, which would matter for anything that reads stderr for diagnostics.

[...]

PR #475, opened 2026-07-30
build: add org.gradle.test-retry to core:ssh module to mitigate upstream sshlib flaky tests

kanazawahere:
This fixes occasional CI flakiness in `core:ssh:testDebugUnitTest` on `SshlibExecContractTest > "stderr and non-zero exit status"`. It's an infrastructure mitigation for a dependency issue, not a real application regression.

- **Root cause:** already documented in `SshlibExec.kt`'s own workaround comment. The `sshlib/cbssh` dependency has a race where `CHANNEL_CLOSE` can cancel the stderr-delivery pump before a slow-to-park consumer reads the buffered output, dropping it.
- **Checked upstream:** `connectbot/cbssh` has no release newer than v0.4.1 (already pinned here), so there's no version bump available yet. The bug is still open on their side.
- **Mitigation:** added the official Gradle test-retry plugin (`org.gradle.test-retry`, v1.6.5), scoped to the `core:ssh` module only. Retries a failed test up to 2 times before failing for real (`maxFailures=20`, `failOnPassedAfterRetry=true`, so a genuinely broken test still fails loudly instead of retrying forever).
- **Verification:** doesn't touch app behavior, only how the test task handles retries. Full CI green (build, test, lint, checks, supply-chain, nightly-abi x64+armv7, release-r8-smoke) before opening this.

Already merged the same commit into our fork (DigitalVersion/Haven) as a sanity check before proposing it here.


GlassOnTin:
Thanks for this — the direction is one I'd already written into #448's next steps ("decide whether to quarantine the assertion meanwhile, so a known upstream defect stops gating releases"), and the dependency-verification and lock-file hygiene here is all correct.

But the config as written doesn't do what it says, and I'd rather you see the evidence than take my word:

**`failOnPassedAfterRetry = true` means "a test that fails and then passes on retry fails the task"** — the plugin's own docs word it as *"Whether tests that initially fail and then pass on retry should fail the task"*, default `false`. So with it set to `true`, the #448 flake gets retried, passes on attempt 2, and CI still goes red. The only net effect is that it now goes red about two test-runs later.

I reproduced it standalone rather than reasoning about it — a JUnit test that fails its first execution and passes every later one, with your exact retry block:

```
maxRetries=2, maxFailures=20, failOnPassedAfterRetry=true   → BUILD FAILED (exit 1)
maxRetries=2, maxFailures=20, failOnPassedAfterRetry=false  → BUILD SUCCESSFUL (exit 0)
```

Same test, same plugin (1.6.5), same Gradle (9.4.1); only that flag differs.

I think the rationale in the description crossed two things over: a *genuinely* broken test fails all its retries and fails the build no matter what this flag says — that's what `maxRetries` governs. `failOnPassedAfterRetry` only decides the flaky-pass case, which is precisely the case you want absorbed.

**Second thing:** `tasks.withType<Test>` applies the retry to every test in `core:ssh`. Since #448 is real output loss rather than test noise, I don't want a *new* flake anywhere else in the module silently absorbed. The plugin's filter scopes it, and I verified that too — with an exact-class filter, the named class is absorbed while a different flaky class in the same module still fails the build:

```kotlin
tasks.withType<Test> {
    retry {
        maxRetries.set(2)
        failOnPassedAfterRetry.set(false)
        filter { includeClasses.add("sh.haven.core.ssh.sshlib.SshlibExecContractTest") }
    }
}
```

(Careful with globs there: I first tried `*FlakyTest` in the probe and it matched a second class I meant to leave out.)

If you make those two changes I'll merge it. If you'd rather not, say so and I'll take it from here — either way the finding belongs on the PR, not in a silent rewrite.

One note for the record, since this mitigates rather than fixes: #448 stays open. The underlying sshlib `CHANNEL_CLOSE`-vs-delivery-pump race can lose real stderr on the experimental engine, and retry-green must not be read as that having gone away.

kanazawahere:
Checked this against the plugin's README before touching anything: "failOnPassedAfterRetry = false (ie. the default for this plugin)". You're right, our first pass had it backwards. Pushed both fixes: `failOnPassedAfterRetry` set to false, and the retry scoped to just `SshlibExecContractTest` via the filter, using your exact snippet. CI is green on the new commit (build, test, lint, checks, supply-chain, nightly-abi x64/armv7, release-r8-smoke, run 30622379438). Agreed on #448 staying open: this is a mitigation, not a fix, and a passing retry doesn't mean the underlying sshlib race is gone.

GlassOnTin:
Thanks for this, and apologies — **this got overtaken by a change on main, and the timing is on my side, not yours.**

You opened this on 30 July. On 2 August I landed `083faf10`, which took the same problem in a different direction: the sshlib contract tests are now excluded from the gating test run and re-run in their own step with `continue-on-error: true`. Your PR was correct when you wrote it.

## What would happen if I merged it today

I checked rather than assumed. On current main, CI's gating job runs:

```
./gradlew testDebugUnitTest :app:testArm64DebugUnitTest -PexcludeSshlibContractTests=true
```

and `core/ssh/build.gradle.kts` turns that flag into `exclude("sh/haven/core/ssh/sshlib/**")`. Running it locally:

| run | test classes | `SshlibExecContractTest` |
|---|---|---|
| `-PexcludeSshlibContractTests=true` | 46 | **absent** |
| without the flag | — | present |

So the class your filter names never executes in the gating job, and the second step that does run it is `continue-on-error: true`, where retrying changes nothing either way. Merged as-is, the plugin would be inert.

## The part worth actually deciding

I don't think your approach is the *worse* one, and I want to put that honestly rather than close this as redundant.

- **What's on main now** removes the flake by removing the gate. A genuine sshlib regression — not the known race, an actual break — would leave CI green, and I'd only catch it by reading a step nobody reads.
- **Yours** keeps the suite gating and absorbs the known flake, scoped to one class with `failOnPassedAfterRetry=false`. A real regression still fails after three attempts.

That distinction matters more than it looks, because of #58: the plan is to flip sshlib from opt-in to default in 6.0.0, and the contract suite is the parity gate for that decision. A gate that cannot fail is not much of a gate — which argues for your version over mine.

If you're up for it, the change I'd take is: drop `-PexcludeSshlibContractTests=true` and the `continue-on-error: true` step from `ci.yml`, keeping your retry config as the thing that absorbs the known race. That restores the gate without restoring the red builds. I'd want the retry to stay scoped to the single class exactly as you have it.

If you'd rather not, that's completely fair — say so and I'll either do it myself crediting this PR, or leave main as it is.

## Independent of all that

Your `verification-metadata.xml` regeneration added `protoc-4.29.3-linux-aarch_64.exe`, which main doesn't have. That's a real portability win for anyone building on an ARM machine and has nothing to do with the retry question — I'll take that regardless of what we decide above.

I went to correct your comment's claim that `failOnPassedAfterRetry=false` is "the plugin's own default", then checked the 1.6.5 jar before typing it. `TestRetryTaskExtensionAdapter` does `getFailOnPassedAfterRetry()` → `iconst_0` → `convention(...)`, so it defaults to `false` and you're right. Mentioning it only because the comment is doing real work there and I'd rather confirm it than leave you wondering whether I'd read it.

GlassOnTin:
Merged — your two commits are on `main` as `4c8b0423` and `e56d954a`, authorship intact. I applied them directly rather than through the button because I wanted a CI comment in the same push; nothing of yours was changed.

**I need to correct my last comment.** I said the plugin would be "inert" merged onto current main. That was wrong, and wrong in the direction that matters.

It's inert in the *gating* run — that part holds, the class is excluded there. But the second step runs `--tests 'sh.haven.core.ssh.sshlib.*'`, which does include `SshlibExecContractTest`, so your retry very much applies to it. I wrote that retrying "changes nothing either way" in a `continue-on-error: true` step. It changes the one thing that step exists to produce: its verdict. That comment already told the reader to distinguish "failing every run" from "flaked once" — and gave them no way to do it. Three attempts is exactly that mechanism.

I verified it rather than reasoning about it a second time. Added a deliberately failing test to `SshlibExecContractTest`, and another to `ConnectionConfigTest`:

| class | testcase entries in the XML |
|---|---|
| `SshlibExecContractTest` | **3** — retried |
| `ConnectionConfigTest` | **1** — not retried |

Both still failed the build (exit 1), so a genuinely broken test is not being swallowed. Your scoping is exactly right, and `maxRetries=2` with `failOnPassedAfterRetry=false` does what the comment says.

Full `core:ssh` suite after applying: 454 tests, 0 failures.

**On re-gating — I've talked myself out of my own suggestion.** I proposed dropping `-PexcludeSshlibContractTests=true` so the suite would gate again. Re-reading the rationale I'd written into `ci.yml`, the exclusion is a product decision rather than an accident: sshlib is the opt-in preview engine, and an upstream race in it shouldn't block shipping fixes to the JSch engine that users are actually running. Re-gating would trade a real cost for a theoretical one. What was actually missing was a trustworthy signal from the non-gating step, and that's what you've supplied.

So no follow-up work needed from you. The `protoc-4.29.3-linux-aarch_64.exe` hash came along with it too.

Thanks for the patience on this one — it sat for four days, got overtaken by a change of mine, and then I mis-diagnosed it. The PR was right.

GlassOnTin:
Landed on main as 4c8b0423 + e56d954a — closing since the commits are in, not because the change was rejected.

PR #481, opened 2026-08-01
Connect Deeplink: Resolve Ambiguous Matches & Implement Warm-Start Trampoline

### LINKED ISSUE #305: Feature: `haven://connect` deep link to launch a connection (optional startup command)
Haven already handles one deep link — `haven://renew-cert/<keyId>` (parsed in `MainActivity.handleRenewCertDeepLink`, declared in `app/src/main/AndroidManifest.xml`).

Would you accept a sibling **`haven://connect`** that launches a connection from URI params, e.g.:

```
haven://connect?transport=mosh&user=<u>&host=<h>&port=<p>&session=<s>
```

When `session` (or a `cmd`) is present, run it as a startup command after connect — e.g. `tmux attach -t <s>`.

**Use case:** a self-hosted dashboard lists tmux sessions; tapping a session card should drop me straight into that session over Mosh in Haven, in one tap. Haven already has the hard parts (Mosh transport + tmux-aware session restore) — it's just missing a parametrized link surface to launch into a specific host/session.

**Question on shape:** does `ConnectionConfig` (core/ssh) already have a startup-/initial-command hook that a Mosh session runs after connect, or should one be added? That decides whether this is mostly an intent-filter + a `handleConnectDeepLink` (mirroring the renew-cert handler) or also a small addition to the session layer.

Happy to send a PR if you greenlight the shape (scheme name, param set, where the startup command should live). Thanks for Haven — the multi-transport + Mosh + tmux-restore combo is exactly what I want on mobile.

kanazawahere:
# Connect Deeplink: Resolve Ambiguous Matches & Implement Warm-Start Trampoline

This PR addresses issue #305 by introducing an explicit connection profile `id` parameter to the `haven://connect` deep link and implementing a clean trampoline activity to support warm-start deep linking without creating redundant task/panes.

## Problem Description

Previously, the deep link pattern was matched solely by connection properties (`host`, `user`, `port`, `transport`). However, when multiple saved connection profiles share the same host/user/port (such as multiple Seer card tmux sessions running on a single host), the deep link matching becomes ambiguous. The app would fallback to opening a blank "New Connection" screen instead of connecting.

## Solution

1. **Explicit ID Parameter**: Added support for an optional `id` parameter (the connection profile's UUID) to the deep link:
   ```
   haven://connect?id=<uuid>
   ```
   If provided, the app will match the profile uniquely by its ID, bypassing the ambiguity on host/user.
2. **Warm-Start Trampoline**: Routed deep links via a clean `DeepLinkActivity` trampoline to dispatch to the active `MainActivity` instance. This prevents launching redundant instances of `MainActivity` and ensures the deep link runs smoothly when the app is already open (warm-start).
3. **Graceful Fallback**: Retained host-matching logic as a backward-compatible fallback when `id` is missing.

---

## Test Verification

Verified all 4 test cases successfully on a real Android device (`seeker`):

| Test Case | Description | Result | Screenshot |
|---|---|---|---|
| **Case A** | **Cold-start**: App is forced stopped, then deep link is opened. | **Passed**. App cold-starts and shows the "Confirm Connect" dialog targeting the exact profile. | Case A Screenshot |
| **Case B** | **Warm-start**: App is already open, then deep link is opened. | **Passed**. App focuses and handles intent inline without recreating screens/tabs. | Case B Screenshot |
| **Case C** | **`id=` exact match**: Opens deep link with unique profile ID. | **Passed**. Correctly targets the `seer · video-gen tiếp` profile even though there are 20+ other profiles sharing the same host. | Case C Screenshot |
| **Case D** | **No `id=` (regression)**: Opens deep link with only host parameter. | **Passed**. Connects to the single matching host or opens New Connection editor pre-filled without crash. | Case D Screenshot |

Closes #305


kanazawahere:
Closing — this was opened from the fork's `main` branch by mistake, so it dumps ~55 unrelated files (CI changes, DB migration, an in-progress browse-mode feature that's fork-only) under a title that only describes the deep-link fix. Replacing with a properly-scoped PR built on top of `upstream/main`.

PR #483, opened 2026-08-01
fix: route deep links through trampoline DeepLinkActivity to active MainActivity

kanazawahere:
### Root Cause
When `MainActivity` is in a warm-start state (running in the background), incoming deep links can trigger the Android system's internal task management with flags like `START_TASK_TO_FRONT`. In certain lifecycle flows, Android brings the task containing `MainActivity` to the front but fails to redeliver the new intent to its `onNewIntent` callback, causing the deep link parameters (e.g. connection ID) to be lost or ignored.

### Solution
We introduce a lightweight trampoline activity `DeepLinkActivity` to capture all deep link intents:
1. `DeepLinkActivity` is declared with `excludeFromRecents="true"`, `taskAffinity=""`, and a translucent theme.
2. It tracks the active `MainActivity` instance using a static `WeakReference` (`MainActivity.getActiveInstance()`).
3. If `MainActivity` is alive and active, `DeepLinkActivity` immediately passes the deep link URI to it via `activeActivity.handleDeepLinkUri(data)`.
4. It then issues a forward Intent to `MainActivity` with explicit flags (`FLAG_ACTIVITY_NEW_TASK or FLAG_ACTIVITY_CLEAR_TOP or FLAG_ACTIVITY_SINGLE_TOP`) to ensure standard intent delivery before finishing itself.


kanazawahere:
Closing — same root cause as #481: this branch was cut from the fork's `main` (which has merged fork-only features upstream doesn't have yet), so the diff carries ~49 unrelated files. Replacing with a clean PR containing only the 3 commits that actually implement this fix, rebased onto `upstream/main`.

PR #486, opened 2026-08-01
Fix deep link profile matching and warm-start support

### LINKED ISSUE #305: Feature: `haven://connect` deep link to launch a connection (optional startup command)

kanazawahere:
This PR resolves two deep link issues:

### (a) Fix profile matching when host/user/port are identical (Fixes #305)
- Adds an `id=` parameter to correctly match profiles even when the host, user, and port are identical.

### (b) Fix deep link warm-start support
- Implements a `DeepLinkActivity` trampoline and a static `WeakReference` callback to `MainActivity` to ensure deep links function correctly when the app is already open (warm-start).
- Verified on a physical Seeker device for both warm-start and cold-start scenarios.

This is a clean pull request containing only the 5 files related to deep link logic, replacing #481 and #483 which were closed due to accidental pollution from the fork's main branch.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -53,6 +53,17 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <meta-data
+                android:name="android.app.shortcuts"
+                android:resource="@xml/shortcuts" />
+        </activity>
+
+        <activity
+            android:name=".DeepLinkActivity"
+            android:exported="true"
+            android:excludeFromRecents="true"
+            android:taskAffinity=""
+            android:theme="@android:style/Theme.Translucent.NoTitleBar">
             <!-- step-ca renewal notification deep-link: tapping the
                  "cert expiring" notification fires haven://renew-cert/<keyId>
                  which lands here, then the activity hands the keyId off to
@@ -77,9 +88,6 @@
                     android:scheme="haven"
                     android:host="connect" />
             </intent-filter>
-            <meta-data
-                android:name="android.app.shortcuts"
-                android:resource="@xml/shortcuts" />
         </activity>
 
         <activity
--- a/app/src/main/kotlin/sh/haven/app/ConnectDeepLink.kt
+++ b/app/src/main/kotlin/sh/haven/app/ConnectDeepLink.kt
@@ -26,22 +26,26 @@ object ConnectDeepLink {
         val session: String?,
         /** Optional remote command (Tin attach / clean tmux). `command` or `startupCommand` query. */
         val command: String? = null,
+        val id: String? = null,
     )
 
     /**
      * Build [Params] from a query-parameter getter (e.g. `uri::getQueryParameter`).
-     * Returns null when no `host` is present — a connect link without a host
-     * is a no-op rather than an error.
+     * Returns null when no `host` (and no `id`) is present — a connect link without
+     * a host or id is a no-op rather than an error.
      */
     fun parse(query: (String) -> String?): Params? {
-        val host = query("host")?.trim()?.takeIf { it.isNotEmpty() } ?: return null
+        val host = query("host")?.trim()?.takeIf { it.isNotEmpty() }
+        val id = query("id")?.trim()?.takeIf { it.isNotEmpty() }
+        if (host == null && id == null) return null
         return Params(
-            host = host,
+            host = host ?: "",
             username = query("user")?.trim()?.takeIf { it.isNotEmpty() },
             port = query("port")?.trim()?.toIntOrNull(),
             transport = query("transport")?.trim()?.lowercase()?.takeIf { it.isNotEmpty() },
             session = query("session")?.trim()?.takeIf { it.isNotEmpty() },
             command = (query("command") ?: query("startupCommand"))?.trim()?.takeIf { it.isNotEmpty() },
+            id = id,
         )
     }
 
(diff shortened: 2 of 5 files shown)


GlassOnTin (changes requested):
Thanks — the `id=` parameter is the right fix for #305, and I want that part. The tests you added for it are good. One blocking problem in the warm-start half, and a question that might let you delete most of it.

## Blocking: warm start now handles the link twice

`DeepLinkActivity.onCreate` does both of these:

```kotlin
val activeActivity = MainActivity.getActiveInstance()
if (activeActivity != null) {
    activeActivity.handleDeepLinkUri(data)      // delivery 1
}
startActivity(forwardIntent)                    // delivery 2 → onNewIntent
```

When Haven is already open, both fire. `onNewIntent` calls `handleConnectDeepLink`, which reaches `agentUiCommandBus.emit(...)` — so one `haven://connect` tap emits two connect commands.

The existing guard doesn't catch it, and it's worth seeing why:

```kotlin
val params = ConnectDeepLink.parse { data.getQueryParameter(it) } ?: return
intent.data = null      // ← replay guard
```

`handleDeepLinkUri` builds a **fresh** `Intent().apply { data = uri }` and passes that. So `intent.data = null` clears the synthetic intent, which nothing else ever reads, and leaves the real forwarded intent untouched. The guard is intact and simply isn't on the path.

## The question: is the trampoline needed at all?

`MainActivity` is `android:launchMode="singleTask"`. With singleTask, an intent targeting an existing instance is *always* delivered to `onNewIntent` — Android does not recreate it, and no `CLEAR_TOP`/`SINGLE_TOP` flags are required to make that happen. That's the mechanism your `startActivity` is already relying on.

If that's so, then `getActiveInstance()`, the `WeakReference`, `handleDeepLinkUri` and the `onDestroy` teardown can all go, and the fix is:

```kotlin
class DeepLinkActivity : Activity() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        intent?.data?.let {
            startActivity(Intent(this, MainActivity::class.java).apply {
                action = Intent.ACTION_VIEW
                data = it
                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
            })
        }
        finish()
    }
}
```

That removes the double delivery and the static `Activity` reference in one go.

**But you tested on a device and I have only read the code, so I'd rather ask than assert:** what exactly was failing on warm start before? With singleTask and the filters on `MainActivity`, `onNewIntent` should already have fired. If you saw it genuinely not arrive, that's a different bug and I want to know what it was — it's possible the trampoline is load-bearing for a reason I can't see from here, in which case keep it and just drop the manual callback.

## Smaller things

**Empty host when `id` misses.** `parse` sets `host = host ?: ""`, and `matches` returns nothing when the id matches no profile — so `resolve` falls through to `PrefillNewConnection("", ...)` and opens the editor with a blank host. A stale or mistyped id gives the user an empty form rather than anything actionable. Worth either keeping `host` nullable or short-circuiting an unmatched `id` to something clearer.

**The `narrowCandidates` heuristic.** Narrowing duplicate host/user/port matches by substring-matching `session`/`command` against `prof.remoteCommand` is fuzzy in a way that can pick the wrong profile — a session named `a` substring-matches almost any command. Given `id=` now exists as the exact answer to #305, I'd rather drop the heuristic than carry a rule that silently connects to the wrong host. If you want it kept, it needs a test with two profiles whose commands both contain the session string.

CI being green here isn't evidence against any of the above — the double delivery is lifecycle behaviour with no test covering it.

Happy to take this in two pieces if easier: the `id=` parameter and its tests can land on their own straight away.

GlassOnTin:
Landed — your three commits are on `main` as `0ad49be4`, `bb76eb29` and `c864750f` with authorship intact, plus a follow-up of mine (`9eb232c8`).

**Your commit message answered the question I'd asked, and it changed my recommendation.** I'd suggested deleting the direct-call path, on the reasoning that `MainActivity` is `launchMode="singleTask"` so the platform delivers `onNewIntent` by itself. Then I read `bb76eb29`: *"fixing warm start issue on ColorOS"*. That's the answer — you weren't working around the platform, you were working around a ROM. I can't reproduce that, you verified it on a device, and dropping either path would mean trusting the other on a ROM known for not behaving. (For what it's worth, #494 landed this week and turned out to be the same ROM family cutting a socket while leaving the process alive. It has form.)

So both paths stay, and the handler is idempotent instead: a URI seen again within 3 s is ignored. Whichever delivery arrives first wins. That fixes the double-emit without betting on which mechanism works.

**Also changed:** an `id` that matches nothing now falls back to host matching when the link carried a host. `matches` returned early on a non-null id, so a stale bookmark — profile deleted and recreated keeps its host but gets a new id — ignored the host too and opened a New-Connection editor with a blank field.

**On the narrowing heuristic:** I said in review I'd rather drop it. I kept it, because with `id=` absent from any hand-written link it's the only thing that disambiguates two profiles differing solely by tmux session, which is exactly #305's scenario. It needed the test I asked for, so I wrote both halves:

- `session=play` against commands `tmux new -A -s work` / `tmux new -A -s play` → connects to the right one.
- `session=work` against `work` / `workshop` → **prefills instead of guessing**, since connecting to the wrong host is worse than asking.

Verification: mutating the id fallback reds exactly the one test that covers it, so it isn't decorative. `:app` 287 tests green, `core:ssh` 454 green, preflight clean.

**Not verified by me:** the ColorOS warm-start behaviour itself, and the dedupe window against a real double delivery on hardware. If you get a chance to re-test on your device and the link fires once rather than twice, that would close the last gap — and if 3 s turns out to be too short for the two paths on that ROM, say so and I'll widen it.

Thanks for this one. The `id=` parameter is the right shape for #305, and the trampoline is a genuine finding I'd have argued you out of if you hadn't put the reason in the commit message.

GlassOnTin:
Landed on main as 0ad49be4 + bb76eb29 + c864750f, with follow-up 9eb232c8. Closing because the commits are in.

PR #532, opened 2026-08-13
ci: scan pull requests for credentials and injection with ThreatCrush

ralyodio:
Adds a pull-request workflow that scans the diff for hardcoded credentials,
injection, SSRF and unsafe deserialisation. Results go to the Security tab as
SARIF and to a comment on the pull request.

**It is report-only.** `failOn` is empty, so it annotates and never fails a build.
A repository with pre-existing findings should get a report on its first install,
not a blocked pull request — a gate that fires on everything gets switched off
within a day. Tighten it to `critical,high` in the workflow once any backlog is
triaged.

- `.github/workflows/threatcrush-scan.yml` — the workflow
- `.github/scripts/threatcrush-to-sarif.py` — a compatibility shim for CLI versions
  older than native SARIF output; unused once the installed CLI can emit it itself

Permissions are least-privilege (`contents: read`, `pull-requests: write`,
`security-events: write`). It runs on `pull_request`, not `pull_request_target`,
so contributor code never executes with your secrets in scope. The SARIF upload
is `continue-on-error` and degrades quietly where code scanning is unavailable.

Disclosure: I maintain [ThreatCrush](https://github.com/profullstack/threatcrush).
It is free and MIT, and the workflow installs it from npm — nothing here phones
home. If this is not something you want, closing it is the right answer, and I
will not send another.

### DIFF
--- a/.github/scripts/threatcrush-to-sarif.py
+++ b/.github/scripts/threatcrush-to-sarif.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+"""Convert ThreatCrush terminal output to SARIF 2.1.0.
+
+Compatibility shim for CLI versions older than native ``--format sarif``.
+When the CLI can emit SARIF itself the workflow uses that and never runs this
+file; parsing a human-readable stream is strictly worse and exists only so a
+repository is not left unscanned while waiting for a release.
+
+It **fails closed**. If it cannot recognise the output it exits non-zero and
+dumps what it saw. Emitting empty SARIF instead would report "0 findings",
+which is indistinguishable from a clean scan and is the single most expensive
+thing a security tool can get wrong.
+
+Three details of the format, each of which is load-bearing:
+
+* Severity is bare for ``CRITICAL`` and bracketed for ``[HIGH]``/``[MEDIUM]``/
+  ``[LOW]``. One regex shape misses half the findings.
+* ``File:`` paths are relative to the scan root, not the repository root. Left
+  unprefixed, every finding resolves to nothing in the consumer's view of the
+  repo. Hence ``--path-prefix``.
+* Whole-file findings report line ``:0``. SARIF requires ``startLine >= 1``.
+
+``Code:`` lines are redacted excerpts of the match. They are skipped rather
+than parsed, both because matching them would double-count every finding and
+because a redacted excerpt tells a reader nothing the ``Info:`` line does not.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+
+ANSI = re.compile(r"\x1b\[[0-9;]*[A-Za-z]")
+
+# `   CRITICAL  AWS Access Key`  /  `  [HIGH] Sensitive File`
+SEVERITY_LINE = re.compile(r"^\s*(?:\[(CRITICAL|HIGH|MEDIUM|LOW|INFO)\]|(CRITICAL))\s+(.+?)\s*$")
+FILE_LINE = re.compile(r"^\s*File:\s*(.+?):(\d+)\s*$")
+INFO_LINE = re.compile(r"^\s*Info:\s*(.+?)\s*$")
+
+# Proof that a scan ran to completion. Without one of these we are looking at a
+# crash, a help screen, or an unrecognised release — never at a clean result.
+FOOTER = re.compile(r"^\s*(?:\d+\s+issue\(s\)\s+found|.*No security issues found)")
+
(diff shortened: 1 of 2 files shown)


GlassOnTin:
Thanks for the careful write-up — and credit where due: the token handling is exactly right (`pull_request` rather than `pull_request_target`, with the reasoning documented inline), and report-only is the right default for a first install.

Declining all the same, on one structural ground: the workflow executes `npm install -g "@profullstack/threatcrush@latest"` in CI on every pull request. An unpinned, auto-updating third-party executable is a standing supply-chain surface — whoever can publish that package can run code in this repository's CI from that point on, forever, without a further PR. This project hash-pins every native source tarball and git dependency it builds from (13 pins in the ffmpeg stack alone, verified by tampering); CI tooling has to clear the same bar, and `@latest` structurally can't.

Pinning to an exact version + integrity hash would address that specific objection, but I'll be honest about the second-order cost too: a scanner is a recurring trust relationship, not a one-time diff — each version bump is a re-audit of somebody else's tool against the repo's CI. For a repo this size I'd rather keep that surface at zero than manage it.

Appreciate the interest in the project's security posture — genuinely. If ThreatCrush finds something in the *current* codebase worth reporting, an issue with the finding would be very welcome, and gets acted on faster than a workflow would have surfaced it.

ralyodio:
This is a fair decline and I'm not going to argue it — but the specific defect you named was real, and I'd rather close the loop than let it sit.

You were right about `@latest`, and you weren't the only one to catch it: SonarCloud failed a quality gate on the same line in another repository (`githubactions:S8543`), and CodeRabbit independently scored a third one "Moderate" for handing an unpinned scanner a write-scoped job. Three reviewers, same finding, none of them me. Your framing was the sharpest of the three — a standing grant to whoever can publish the package, with no further PR required — and the ffmpeg-stack comparison is the part that actually landed: it isn't reasonable to ask a project to hold CI tooling to a lower bar than the tarballs it builds from.

Fixed, for whatever it's worth now: the install is pinned to an exact version and runs with `--ignore-scripts`, and checkout runs with `persist-credentials: false`. I pushed that to all 24 requests still open elsewhere rather than only the ones people had objected to.

On the second-order point — that a scanner is a recurring trust relationship and each bump is a re-audit — I don't have a rebuttal, because you're right and pinning doesn't fix it. It converts an unbounded grant into a recurring cost, and "I'd rather keep that surface at zero" is a coherent answer to a recurring cost. That's a reasonable place to land for a project this size.

Thanks for the offer on findings. I won't take you up on it speculatively — a large share of what this reports on a first pass is false positives, and an issue full of unverified findings is worse than no issue. If I ever have something in Haven I've confirmed by hand, I'll send that and nothing else.

No further requests from me here, as promised in the original.

GlassOnTin:
That's about the best possible outcome for a declined PR — the fix pushed to two dozen other repos is worth more than a merge here would have been. And the hand-verified-findings-only standard is exactly right; that offer stays open indefinitely. Thanks for engaging with the reasoning instead of the outcome.

ralyodio:
Closing the loop on one specific thing, and then I'll leave this alone — I said no further requests here and I mean it, so to be explicit up front: **this is not a re-ask, and I'm not asking you to reconsider.**

You wrote that pinning to *"an exact version + integrity hash"* would address the structural objection. The exact version shipped at the time. The integrity hash didn't — I only did half of what you named, which I didn't notice until I went back through this thread.

It's done now. The workflow no longer installs on the registry's word: it downloads the pinned tarball, hashes it, compares the SHA-512 against a value committed in the workflow file, and refuses to install on a mismatch. Verified `npm pack` reproduces the published tarball byte-for-byte, so the comparison is against the registry's own `dist.integrity` rather than a repack of it, and a wrong hash exits non-zero without installing. A mismatch isn't retried, unlike a download failure — a network blip and a registry serving bytes nobody signed off on are different events, and retrying the second just asks again until it succeeds.

**None of that changes your actual reason for declining**, and I want to be straight about that rather than let a fix for the first objection look like an answer to the second. You said a scanner is a recurring trust relationship, each bump is a re-audit, and you'd rather keep that surface at zero than manage it. The hash doesn't touch that — it closes the window on *one* version's bytes, and every bump reopens the same question. I agreed with you then and I still do. It also only covers the CLI tarball, not its dependency tree; `--ignore-scripts` limits what that tree can do at install time, but "limits" isn't "eliminates."

So: your review made the tool better twice over, and the second time you weren't even here for it. The ffmpeg-stack comparison — that CI tooling has to clear the same bar as the tarballs you build from — is now written into the pack's own documentation as the reason the input exists.

The hand-verified-findings-only standard still holds on my side. Thanks again.

PR #534, opened 2026-08-13
fix(terminal): reliably paste uploaded file path after Attach flow

kanazawahere:
## Summary
- The Terminal's Attach → "Send a file" flow uploaded the file via SFTP correctly, but the remote path was silently never pasted into the terminal after returning from the folder picker — no error shown to the user.
- Root cause: the ViewModel called `tab.sendInput(...)` directly from inside the coroutine launched by `runAttachFlow`, racing against Compose recomposition/navigation back to the Terminal screen. This mirrors a pattern that was already fixed correctly for the existing QR/barcode "scan injection" feature (`pendingScanInjection`), but the Attach flow had not been migrated to it.
- Fix: route the attach payload through a `_pendingAttachInjection` StateFlow, consumed by a `LaunchedEffect` in the Compose screen (same shape as `pendingScanInjection`), which looks up the currently active tab, applies bracketed-paste wrapping if enabled, and performs the injection safely post-recomposition.
- Also added fail-loud handling: if no active tab exists when either injection (attach or the pre-existing scan) fires, the user now sees a toast ("No active tab to paste into") instead of the payload silently disappearing — this was a latent gap in the original scan-injection pattern too.

## Test plan
- [x] Manual repro on a physical device (JP1, AOSP) with a CI-built debug APK: Attach → Send a file → pick file → Choose folder → Use this folder. Confirmed the uploaded file's remote path is now pasted at the terminal cursor, wrapped in bracketed-paste when the shell has it enabled.
- [x] Independent code review pass (separate reviewer) confirmed the new code correctly mirrors the existing `pendingScanInjection` pattern, is thread-safe (StateFlow, Main-scoped), and does not leave the pending value stuck.
- [ ] No automated test added for this flow yet — `TerminalViewModelTest.kt` currently has no coverage of `runAttachFlow`. Happy to add one if maintainers want it as a condition for merge.

## Known residual limitation (documented, not blocking)
- If two attach/scan completions happen back-to-back before Compose recomposes, `StateFlow` conflation could theoretically drop the first injection in favor of the second. This is a pre-existing characteristic of the `pendingScanInjection` pattern this fix reuses, not something newly introduced — noting it here for visibility rather than silently shipping it unremarked.


GlassOnTin:
The fix itself is real and wanted — thank you. The Attach flow racing `tab.sendInput` against recomposition is exactly the failure the scan-injection path already solved, and draining a `pendingAttachInjection` StateFlow the same way is the right shape. I verified the mechanism against main and I'd take it as-is.

The problem is the branch it rode in on: this PR carries **48 commits of your fork's whole state** alongside the fix — an aarch64-host build port of `build-ffmpeg`/`rdp-kotlin`, a new `tinder/` preview feature, a Room schema bump to 81, CI changes, and a `spice-kotlin` submodule repointed to a different owner's fork. None of those can merge here (the submodule repoint and the schema bump in particular — the latter is a one-way door we sequence very deliberately), and they bury a ~40-line fix under +5,363.

Two ways forward, your choice:

1. **You open a clean PR** containing just the Attach-flow fix: the `pendingAttachInjection` additions in `TerminalViewModel`, the drain effect in `TerminalScreen`, and ideally a small test beside the existing scan-injection one. (The `` change to the existing scan payload isn't needed — main has a raw ESC byte there, which is equivalent at runtime; keeping the diff to the new flow keeps it reviewable.)
2. **I cherry-pick it for you**, preserving your authorship on the commit, and note you as the fix's author in the changelog.

Say which — and if the aarch64-host build work or the preview feature are things you'd like upstream, each is worth its own issue first so the design conversation happens before the diff. Closing this one either way once the fix lands by one of the two routes.

kanazawahere:
Thanks for the review — opened a clean replacement off upstream/main: #535. Closing this one.

GlassOnTin:
Superseded by #535 (the clean re-PR), which is applied to main with your authorship — closing this one. Thanks again for splitting it out.

PR #535, opened 2026-08-13
fix(terminal): reliably paste uploaded file path after Attach flow

kanazawahere:
Clean replacement for #534, per @GlassOnTin's review — same fix, based straight off `upstream/main` (no fork-state noise this time).

## Summary
- The Attach → "Send a file" flow uploaded correctly via SFTP, but the remote path silently never landed in the terminal — `tab.sendInput(...)` was called directly from inside the coroutine launched by `runAttachFlow`, racing against Compose recomposition/navigation back to the Terminal screen.
- Fix: route the payload through a `pendingAttachInjection` StateFlow drained by a `LaunchedEffect` in the screen — the same shape `pendingScanInjection` already uses for the QR/barcode scan feature — applying bracket-paste the same way.
- Also added a toast (`terminal_paste_no_active_tab`) instead of silently dropping the payload if no active tab exists when either injection fires — a latent gap the existing scan-injection path had too.
- Left the existing scan payload's bracket-paste escape literal untouched, per your note that the raw-byte and `` spellings are equivalent at runtime and not worth touching in this diff.

## Test plan
- [x] Manual repro on a physical device with a debug APK built from this fix: Attach → Send a file → pick file → Choose folder → Use this folder → path now pastes at the terminal cursor, bracket-paste wrapped when the shell has it enabled.
- [ ] No automated test yet for `runAttachFlow` — happy to add one alongside the existing scan-injection test if you'd like it as a condition for merge.

### DIFF
--- a/feature/terminal/src/main/kotlin/sh/haven/feature/terminal/TerminalScreen.kt
+++ b/feature/terminal/src/main/kotlin/sh/haven/feature/terminal/TerminalScreen.kt
@@ -409,8 +409,32 @@ fun TerminalScreen(
                 text
             }
             tab.sendInput(payload.toByteArray())
+        } else {
+            viewModel.showNewTabMessage(context.getString(R.string.terminal_paste_no_active_tab))
         }
         viewModel.consumeScanInjection()
+    }
+
+    // Inject the uploaded file's remote path at the cursor on the active
+    // tab once the Attach → "Send a file" flow finishes, honouring
+    // bracket-paste the same way pendingScanInjection does above. Drains
+    // the StateFlow via consumeAttachInjection so a second attach doesn't
+    // fire on recomposition.
+    val pendingAttachInjection by viewModel.pendingAttachInjection.collectAsState()
+    LaunchedEffect(pendingAttachInjection) {
+        val text = pendingAttachInjection ?: return@LaunchedEffect
+        val tab = viewModel.tabs.value.getOrNull(viewModel.activeTabIndex.value)
+        if (tab != null) {
+            val payload = if (tab.bracketPasteMode.value) {
+                "[200~$text[201~"
+            } else {
+                text
+            }
+            tab.sendInput(payload.toByteArray())
+        } else {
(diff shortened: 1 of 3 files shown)


GlassOnTin:
Reviewed and applied — exactly what was asked for, delivered fast.

Verified on my side: both bracket-paste literals carry the real ESC byte (it's invisible in the web diff, so I checked the applied bytes), the drain shape matches the scan-injection pattern, and routing the no-tab case through the existing message channel was the right call — including adding it to the scan path, which had the same silent-drop gap.

Your commit is applied to `main` **as-is with your authorship** via `git am`, with one follow-up commit from me adding the 11 locale translations for `terminal_paste_no_active_tab` (a CI gate requires every key in every locale — not something external contributors should have to carry). Terminal test suite is green on top.

One mechanical note: a crash-fix release is mid-pipeline right now, so the push lands right after it tags (pushing sooner would cancel its CI). I'll close this PR when the commits are on `origin/main` — your fix rides the next release after v5.87.20, with credit in the changelog.

GlassOnTin:
Landed on `main` — closing as merged-in-spirit (the commits were applied via `git am`, so GitHub won't mark the PR merged, but the history tells the true story):

- 6453b0a0 — your fix, your authorship, as-is
- a2443125 — the 11-locale translations for `terminal_paste_no_active_tab`
- 68e0c1a4 — a bonus your PR earned: the red `lint` job on your branch was **not your code**. `settings_terminal_opacity_hint` ("Below 100% Haven…") has been tripping `StringFormatInvalid` on every PR since v5.60.4 — lint reads "% H" as a broken format conversion. Your PR was what surfaced it; it's now marked `formatted="false"` in all locales, so the next contributor gets a green lint job.

CI is green on all three commits ([run 31713481179](https://github.com/GlassHaven/Haven/actions/runs/31713481179)). v5.87.20 (the crash-fix release) is mid-pipeline now; your fix ships in the release after it, credited in the changelog.

Thanks again — clean work, fast turnaround.

PR #561, opened 2026-08-18
fix(i18n): polish Korean translations and restore missing content

Nergis0318:
## 요약
- 기존 한국어 번역의 부자연스럽거나 잘린/오래된 번역을 수정하고 누락된 내용을 복원합니다.
- 키 누락 없음(14개 모듈 `values-ko` 모두 존재). 플레이스홀더/복수형 일치.

## 수정 내용
- **feature/connections**: `helper_ssh_options` 잘림 복구(알고리즘 목록 상세 설명), `auth_methods_description` 잘림 복구, `fido_waiting_body` 잘못된 내용 교체, `password_keyboard_warning` 예시 키보드(HeliBoard 등) 복구, VM 설치 단계 8개 복구(Pixel 9 잔재 제거, 번호/상세 설명 복원)
- **feature/settings**: `verbose_logging_subtitle` RDP 프레임 타이밍 누락 복구, `unpair_message` 어색한 표현 다듬기

## 검증
- 14개 모듈 `values-ko/strings.xml` XML 파싱 성공
- `translatable="false"` 2건(`tasker_blurb`, `keys_discover_cred_detail`) 제외 키 1:1 일치
- `%1$s` 등 플레이스홀더 일치


GlassOnTin:
Thank you — this is a careful piece of work, and I checked it rather than taking the summary on trust. Everything you claimed holds up:

- Both files parse as XML.
- Key parity with the English source: no keys added or dropped.
- **Placeholder parity: zero mismatches** across every string in both modules — the failure mode I worry about most in a translation change, because `%1$s` going missing is a crash at runtime rather than a wrong word.
- No unescaped apostrophes.
- The "restored content" is real, not padding. `settings_verbose_logging_subtitle` in English mentions RDP and frame timings and the Korean had fallen behind; `HeliBoard` appears in the English `password_keyboard_warning` and was absent from the Korean before this PR.

## The one thing blocking it

CI's `checks` job fails on a single step: **"Check translation export is current"**. `docs/i18n/strings.json` is generated from the `strings.xml` files, and it has to be regenerated and committed alongside them:

```
python3 scripts/i18n_export.py
git add docs/i18n/strings.json
git commit --amend --no-edit   # or a follow-up commit, whichever you prefer
git push --force-with-lease
```

I ran it against your branch: it produces exactly 20 changed lines, matching your 20 changed strings, and nothing else. So this is purely the mechanical step, not a problem with the translations.

Nothing else in the run is a real failure — the `test`/`proot-l2s-test`/`release-r8-smoke` jobs show as cancelled because the `checks` failure stopped the run, not because they found anything.

Happy to push that regeneration to your branch myself if you have "allow edits from maintainers" enabled and would rather not bother — say the word. Otherwise push it and I will merge.

GlassOnTin:
Merged to main as b615f5148 (your commit, authorship preserved) plus ff9502aba, the regenerated translation export that the checks job wanted — that regeneration was the only thing standing between this PR and green, and it has to be run on the receiving tree, so I applied your patch directly rather than round-tripping the fork branch.

The changes will be in the next release. Thank you for the careful work; the placeholder and key parity held up exactly as you described.

GlassOnTin:
Released in v5.87.36: https://github.com/GlassHaven/Haven/releases/tag/v5.87.36

Thanks again for the polish; the full set is in this build.

Issue #580, opened 2026-08-23
[Bug] Touch input is unusable in Herdr tabs on Android – incorrect coordinate mapping


binhdnguyen:
https://github.com/user-attachments/assets/4a67b799-67f5-44e4-80d2-45122affd8c6

Description

Touch input does not work correctly when using Herdr inside Haven on Android.

All Herdr tabs and panes are affected. The touch/click position is mapped to the wrong coordinates, making the Herdr interface effectively unusable.

This happens consistently and is reproducible 100% of the time.

The remote display itself appears to be rendered correctly, but the touch coordinates do not correspond to the visible UI position.

Environment

- Haven version: 5.87.46
- Device: Vivo X Fold5
- Android: Android 16
- Architecture: ARM64
- Application affected: Herdr

Steps to reproduce

1. Open Haven on Android.
2. Connect to the remote machine using SSH.
3. Open Herdr.
4. Open or switch between Herdr tabs/panes.
5. Tap on tabs, panes, or other UI elements.
6. The click/touch is registered at a different position from the location that was tapped.

Expected behavior

The pointer/touch position on the remote terminal should exactly correspond to the position tapped on the Android screen.

Actual behavior

Touch coordinates are mapped incorrectly across all Herdr tabs and panes.

The problem is not intermittent. It happens on every tab/session I tested and makes Herdr practically unusable using touch input.

Additional observations

This may be related to coordinate transformation between:

- Android display size
- Haven terminal viewport
- remote terminal dimensions
- Android display scaling
- system insets
- foldable display dimensions/aspect ratio

The displayed UI appears correctly positioned, but the touch-coordinate transformation seems to use a different scale and/or offset.

I can provide a screen recording with Android "Show taps" enabled if needed.

GlassOnTin:
Thanks. That is a well-written report. The observation that the display renders correctly while the touch lands elsewhere is the useful part. It indicates that the problem is in the transform, not the rendering.

## Where this can go wrong

Haven converts a touch into a terminal cell by dividing by the character width and height, then reports that cell to the application. There are two coordinate regimes, and which one is in use depends on the app:

- A full-screen app, or one with mouse tracking active, gets the PTY resized to fit above the keyboard. The rendered content and the touch coordinates then share one origin.
- A plain shell keeps its full height, and the *render* is shifted up instead. Hit-testing adds that shift back.

[...]

PR #587, opened 2026-08-24
Fix IME input dropping repeated control bytes

binhdnguyen:
## Summary

Preserve repeated terminal control bytes in `InputCoalescer` while keeping
the existing IME double-fire workaround for duplicate printable ASCII input.

This fixes incorrect Vietnamese text composition with Gboard in Haven's SSH
terminal, where composed text could appear duplicated, for example:

`độc lập` -> `đôộc lâập`

## Observed behavior

I initially compared the same Vietnamese Gboard input in two Android terminal
apps.

In Moshi, Gboard produced the expected text correctly:

`độc lập`

In Haven, however, Vietnamese composition produced a characteristic duplication
pattern, for example:

`đôộc lâập`

Visually, it looked as if Gboard was replacing an existing character sequence,
but Haven was failing to fully remove the old characters before inserting the
replacement text.

Because of that behavior, the initial suspicion was an Android
IME/InputConnection issue in Haven, possibly around `replaceText()` handling.

However, further instrumentation showed that Gboard, `InputConnection`,
`ImeInputView`, and `KeyboardHandler` were all producing the correct replacement
sequence.

The actual loss happened later in the terminal input pipeline.

## Root cause

Vietnamese Gboard performs composition by replacing previously entered
characters.

For example, while composing:

`đôc` -> `độc`

Gboard replaces the trailing `ôc` with `ộc`.

Haven correctly generated two consecutive DEL (`0x7F`) bytes to remove the two
existing characters, followed by the UTF-8 bytes for the replacement text.

The intended terminal input was therefore:

`7F 7F` + `ộc`

Instrumentation confirmed that both DEL events reached `KeyboardHandler`.

However, `InputCoalescer` contained a historical IME double-fire workaround
that deduplicated any batch containing exactly two identical single-byte inputs.

As a result:

`7F 7F` -> `7F`

Only one DEL reached the remote SSH PTY.

Instead of removing both `ô` and `c`, the terminal removed only `c`, leaving:

`đô`

Then the replacement text `ộc` was appended:

`đô` + `ộc` -> `đôộc`

The same underlying behavior could produce other duplicated Vietnamese
composition results such as:

`lâập`

So the original symptom resembled an IME/InputConnection replacement bug, but
the actual root cause was the byte deduplication heuristic in
`InputCoalescer`.

## Fix

Restrict the historical IME double-fire deduplication heuristic to exactly two
identical printable ASCII bytes (`0x20..0x7E`).

Repeated control bytes are now preserved.

This means input such as:

`aa` -> `a`

continues to use the existing IME double-fire workaround, while legitimate
terminal control input such as:

`7F 7F` -> `7F 7F`

is no longer modified.

This also preserves repeated control bytes such as DEL, TAB, CR, ESC, and
Ctrl-C.

## Verification

Verified on a physical Android device using Gboard Vietnamese over an SSH
terminal session.

Before the fix, while composing `độc`, the remote PTY received:

`64 7F C4 91 6F 7F C3 B4 63 7F E1 BB 99 63`

The final replacement sequence contained only one DEL before `ộc`.

After the fix, the remote PTY receives:

`64 7F C4 91 6F 7F C3 B4 63 7F 7F E1 BB 99 63`

The two consecutive DEL bytes are now preserved correctly.

In particular, the final part changed from:

Before:

`63 7F E1 BB 99 63`

to:

After:

`63 7F 7F E1 BB 99 63`

The second `0x7F` now reaches the remote SSH PTY correctly, and Vietnamese
Gboard composition produces the expected text instead of duplicated output.

## Tests

Added regression tests covering:

- two consecutive DEL bytes are preserved
- repeated control bytes are preserved
- duplicate printable ASCII still uses the existing workaround
- different printable bytes are preserved
- three identical printable bytes are preserved

Full terminal unit test suite:

```text
./gradlew :feature:terminal:testDebugUnitTest

BUILD SUCCESSFUL

### DIFF
--- a/feature/terminal/src/main/kotlin/sh/haven/feature/terminal/TerminalViewModel.kt
+++ b/feature/terminal/src/main/kotlin/sh/haven/feature/terminal/TerminalViewModel.kt
@@ -169,18 +169,42 @@ internal class EmulatorWriteBuffer(
 }
 
 /**
- * Coalesces rapid single-byte inputs into a batch, then deduplicates only
- * the exact IME double-fire pattern (buffer == [X, X]).
+ * Applies the historical IME double-fire workaround to a batch of
+ * single-byte terminal input.
  *
- * Android IMEs often fire both commitText and sendKeyEvent for the same
+ * Only duplicate printable ASCII bytes are eligible for deduplication.
+ * Control bytes must be preserved because repeated control characters
+ * can be intentional terminal input. In particular, Vietnamese Gboard
+ * replacement can legitimately emit two consecutive DEL (0x7F) bytes.
+ */
+internal fun normalizeCoalescedInput(buffer: List<Byte>): ByteArray {
+    val isImeDedupCandidate =
+        buffer.size == 2 &&
+            buffer[0] == buffer[1] &&
+            (buffer[0].toInt() and 0xFF) in 0x20..0x7E
+
+    return if (isImeDedupCandidate) {
+        byteArrayOf(buffer[0])
+    } else {
+        buffer.toByteArray()
+    }
+}
+
+/**
+ * Coalesces rapid single-byte inputs into a batch and applies the historical
+ * IME double-fire workaround to exactly two identical printable ASCII bytes.
+ *
+ * Android IMEs may fire both commitText and sendKeyEvent for the same
  * keystroke, causing onKeyboardInput to be called twice with the same byte.
  * Both calls happen within a single Handler message, so a posted Runnable
  * flushes after all input from the current message is processed.
  *
- * The IME double-fire signature: exactly 2 identical bytes in the buffer.
- * Paste "aa" also matches this, but that's a rare edge case compared to
- * the constant double-fire on every keystroke. Longer paste sequences
- * (e.g., "aab", "43339") are preserved correctly.
+ * Repeated control bytes are intentionally preserved. They can represent
+ * legitimate terminal input; for example, Vietnamese Gboard replacement may
+ * emit two consecutive DEL (0x7F) bytes before sending replacement text.
+ *
+ * Paste "aa" still matches the printable-ASCII heuristic, while longer input
+ * sequences such as "aab" and "43339" are preserved.
  *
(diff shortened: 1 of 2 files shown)


GlassOnTin:
Merged. This is a good find and a well-made change.

I verified it rather than taking the description on trust. Your tests pass, the whole terminal module is green at 140 tests, and I checked the tests can actually fail: reverting the condition to the old "any doubled byte" heuristic fails exactly the two control-byte tests and leaves the printable-ASCII one passing. That is the right shape — it means the tests are pinned to the behaviour that changed and not just to the code as written.

The part of your report that made this reviewable was the byte trace. Two runs of the same composition with and without the fix, showing 7F 7F where there had been a single 7F, leaves nothing to argue about. It also explains why this looked like an InputConnection bug for so long: the loss happens well after the IME layer, so every instrumentation point upstream shows correct data.

Pulling normalizeCoalescedInput out as a pure function was the right call too. The old heuristic was buried inside a synchronized block, which is why it had no tests.

One thing I want to name rather than leave implicit, since it is now the only remaining case: two identical printable ASCII bytes in one batch are still collapsed, so typing or pasting "aa" fast enough to land in a single batch still loses a character. Your change makes that strictly narrower than it was and it is not a regression, but it is still a real edge. The proper fix is to stop guessing and tell a real double-fire from genuine repeat input at the IME layer, where the information exists — which is a bigger change than this one and should not have held yours up. If you want to take that on I would review it; if not I will pick it up separately.

Thanks for the fix, and for the diagnosis on #580 before it.

GlassOnTin:
This shipped in v5.87.57: https://github.com/GlassHaven/Haven/releases/tag/v5.87.57

Thanks for the fix. Narrowing the issue to two DELs in a row surviving the coalescing dedup was the right approach. Keeping the guard to printable bytes rather than dropping it preserved the Gboard behaviour it was there for.

F-Droid will pick it up on its next build cycle. If you are installing from there, it will be a while behind the GitHub release.

Issue #99, opened 2026-04-18
english input method correction


mio-19:
After adding input method support https://github.com/GlassOnTin/Haven/issues/96, the experience with english input method is not optimal.
I am using gboard with haven 5.20.1.
gboard is now giving suggestions on the top bar. there are two cases
+ suggestion's beginning matches what I entered : work as expected
+ I entered a word. gboard suggested a correction : I clicked the correction suggestion, but the text in terminal is not corrected. https://github.com/GlassOnTin/Haven/issues/96#issuecomment-4267837328

GlassOnTin:
Partial update, and honest about what we found.

I spent a few hours instrumenting and iterating on this today. Logcat reveals the exact shape of the problem: Gboard queries our input field **once** at session start (`getTextBeforeCursor(600, 1) → ""`), decides our custom InputConnection isn't a full editor, and then dispatches every keystroke as a separate `commitText("t", 1)` → `commitText("e", 1)` → … — never using the composition protocol (`setComposingText` / `setComposingRegion`) that autocorrect-replacement relies on. Because Gboard never enters composition mode for Haven's field, it also never emits a `setComposingRegion(0, 3) + commitText("the", 1)` pair when you tap a correction. Taps just become no-ops.

Things I tried to flip Gboard into composition mode, none of which worked:

- `fullEditor=true` on `BaseInputConnection` so Gboard has a real Editable to track
- `TYPE_TEXT_FLAG_AUTO_CORRECT` on the inputType
- Dropping `IME_FLAG_NO_EXTRACT_UI` in Standard keyboard mode
- Preserving the editable across commits (previously we cleared on every keystroke, wiping Gboard's context)
- Only clearing editable on `KEYCODE_ENTER` (real command boundary) instead of every key
- A `setComposingRegion` → backspace-then-commit handler, ready to work the moment Gboard ever uses composition

[Termux takes the same position](https://github.com/termux/termux-app/blob/master/app/src/main/java/com/termux/app/terminal/TermuxActivityRootView.java): they deliberately use `TYPE_TEXT_FLAG_NO_SUGGESTIONS` and don't attempt IME autocorrect in the terminal, because custom-View InputConnection + Gboard just doesn't route composition-based corrections the way a real EditText does. That matches what I measured today.

What did ship from this investigation:

[...]

Issue #525, opened 2026-08-09
F-Droid build seems to be broken


connesc:
The F-Droid channel seems to be stuck at v5.86.40.

This seems to come from a broken build, as show in F-Droid logs: https://monitor.f-droid.org/builds/log/sh.haven.app/7631

GlassOnTin:
Thank you — and the log link is what made this diagnosable in minutes rather than days. You are right that it is a broken build rather than a queue backlog, and I had that wrong: I have been checking F-Droid's *published index* today and concluding "F-Droid just lags", when I could have checked the build monitor you linked and seen it failing. That is on me.

## What broke

```
fdroidserver.exception.FDroidException: Found invalid versionCodes for some apps
```

on `sh.haven.app:7631` (v5.87.0), at **2026-08-08 22:43**.

That error is fdroidserver refusing metadata containing **two entries with the same versionCode**. F-Droid's auto-update bot had appended a second, byte-identical 5.87.0 block to our recipe, so versionCode 7631 appeared twice. The build server would not touch the app at all in that state — which is exactly why the channel froze rather than simply falling behind.

## Where it stands now

The duplicate is **already gone**. I spotted it while retargeting our recipe MR and removed it; that merged at **22:53 on 2026-08-08** — about ten minutes after the failed build, which is why the log still shows the broken state.

Current metadata, checked just now against `fdroid/fdroiddata` master:

- 60 version entries, **no duplicate versionCodes**
- 7631 no longer exists (that entry was retargeted), so it will not be retried
- **7651 (v5.87.2)** and **7661 (v5.87.3)** are both present and have **not been attempted yet** — no build log exists for either

So there is nothing left to fix on our side, and nothing for a maintainer to do. The next build cycle should pick up 7661.

## What you will get when it does

Worth saying, because the gap is unusually wide — 5.86.40 to 5.87.3 is a lot, and one of them matters:

**v5.87.4 fixes a privacy bug** (#518) where hostnames, usernames and — through the clipboard path — anything copied in a remote session was written to Android's log in plain text. If you have shared a Haven log with anyone while on 5.86.40, it is worth checking what was in it. That version is not yet in the metadata; it will arrive via the same bot on its next run.

I will keep an eye on the build monitor rather than the index from now on. If it fails again on 7661, please do exactly what you did here — the log link is the useful part.

GlassOnTin:
[...]

PR #589, opened 2026-08-25
fix(terminal): distinguish IME double-fire from genuine repeated input

binhdnguyen:
Follow-up to #587.

#587 correctly stopped collapsing repeated control bytes (Vietnamese Gboard `7F 7F`). The remaining printable-ASCII heuristic still dropped a genuine `"aa"` that landed in one coalescer batch.

This stops guessing at `InputCoalescer`. Gboard's `commitText` + raw-key echo is dropped in `ImeInputView`, where the callback identity exists. `normalizeCoalescedInput` is now a passthrough.

While testing Vietnamese Gboard over SSH, two more IME/terminal desyncs showed up and are included:

- Backspace deleted on the PTY but left the InputConnection `Editable` stale, so Telex kept composing against `getTextBeforeCursor` ghosts. User DEL now shrinks the Editable by one code point (#99: the whole buffer is still not wiped).
- Hold-backspace died after a syllable because `restartInput` tore down Gboard's repeat. Backspace no longer restarts the IME; if Gboard goes silent while DEL is still down, Haven continues sending DEL until `ACTION_UP`.

## termlib
Most of the IME-layer work lives in the `termlib` submodule:

https://github.com/GlassOnTin/termlib/compare/main...binhdnguyen:termlib:fix/gboard-ime-echo-and-backspace

The submodule URL on this branch temporarily points at `binhdnguyen/termlib` so CI can fetch SHA `8a9a689`. After the companion termlib PR merges, that URL can go back to `GlassOnTin/termlib` and the SHA can be retargeted.

## Tests
- `./gradlew :feature:terminal:testDebugUnitTest --tests sh.haven.feature.terminal.InputCoalescerTest` — 5/5 (`[a,a]` preserved; `7F 7F` and other controls still pass through)
- termlib `ImeInputViewTest` — 90/90
- Physical Android device with Gboard Vietnamese over SSH (echo `"aa"`, Telex after delete, hold-backspace)

## Tests that fail if reverted
- `genuine repeated printable ASCII is preserved` fails if the old `[X,X]→[X]` heuristic returns
- `testTypeNullRawViewEventAfterCommitTextIsEchoSuppressed` fails if the echo is delivered twice
- `testTypeNullRawKeyAfterEchoDrainIsGenuineRepeat` fails if a later keystroke is treated as echo

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "termlib"]
 	path = termlib
-	url = https://github.com/GlassOnTin/termlib.git
+	url = https://github.com/binhdnguyen/termlib.git
 	branch = main
 [submodule "mosh-kotlin"]
 	path = mosh-kotlin
--- a/feature/terminal/src/main/kotlin/sh/haven/feature/terminal/TerminalViewModel.kt
+++ b/feature/terminal/src/main/kotlin/sh/haven/feature/terminal/TerminalViewModel.kt
@@ -169,42 +169,27 @@ internal class EmulatorWriteBuffer(
 }
 
 /**
- * Applies the historical IME double-fire workaround to a batch of
- * single-byte terminal input.
+ * Flushes a coalesced single-byte input batch.
  *
- * Only duplicate printable ASCII bytes are eligible for deduplication.
- * Control bytes must be preserved because repeated control characters
- * can be intentional terminal input. In particular, Vietnamese Gboard
- * replacement can legitimately emit two consecutive DEL (0x7F) bytes.
+ * IME double-fire (commitText plus a raw view key) is suppressed in
+ * ImeInputView, not here. This function must not drop repeated printable
+ * ASCII or repeated control bytes: genuine "aa", paste of "aa", and
+ * Vietnamese Gboard replacement DELs (0x7F 0x7F) are all intentional.
  */
 internal fun normalizeCoalescedInput(buffer: List<Byte>): ByteArray {
-    val isImeDedupCandidate =
-        buffer.size == 2 &&
-            buffer[0] == buffer[1] &&
-            (buffer[0].toInt() and 0xFF) in 0x20..0x7E
-
-    return if (isImeDedupCandidate) {
-        byteArrayOf(buffer[0])
-    } else {
-        buffer.toByteArray()
-    }
+    return buffer.toByteArray()
 }
 
 /**
- * Coalesces rapid single-byte inputs into a batch and applies the historical
- * IME double-fire workaround to exactly two identical printable ASCII bytes.
+ * Coalesces rapid single-byte inputs into one SSH write.
  *
- * Android IMEs may fire both commitText and sendKeyEvent for the same
- * keystroke, causing onKeyboardInput to be called twice with the same byte.
- * Both calls happen within a single Handler message, so a posted Runnable
- * flushes after all input from the current message is processed.
+ * A posted Runnable flushes after the current Handler message, so a
+ * burst of 1-byte onKeyboardInput callbacks (IME paste iteration,
+ * per-codepoint dispatch) becomes a single sink() call.
  *
- * Repeated control bytes are intentionally preserved. They can represent
- * legitimate terminal input; for example, Vietnamese Gboard replacement may
- * emit two consecutive DEL (0x7F) bytes before sending replacement text.
- *
- * Paste "aa" still matches the printable-ASCII heuristic, while longer input
- * sequences such as "aab" and "43339" are preserved.
+ * Duplicate printable input is no longer guessed here. Gboard's
+ * commitText+raw-key echo is dropped in ImeInputView where the
+ * callback identity still exists.
  *
  * Multi-byte inputs (toolbar, escape sequences) bypass coalescing.
  */
@@ -226,9 +211,8 @@ private class InputCoalescer(private val sink: (ByteArray) -> Unit) {
             buffer.add(data[0])
         }
 
-        // Post flush to run after the current message completes.
-        // Both IME double-fire calls and paste iteration happen within
-        // one message, so the flush sees the complete batch.
+        // Post flush after the current message so a burst of 1-byte
+        // callbacks is written as one sink() call.
         handler.removeCallbacks(flushRunnable)
         handler.post(flushRunnable)
     }
--- a/feature/terminal/src/test/kotlin/sh/haven/feature/terminal/InputCoalescerTest.kt
+++ b/feature/terminal/src/test/kotlin/sh/haven/feature/terminal/InputCoalescerTest.kt
@@ -24,7 +24,7 @@ class InputCoalescerTest {
     }
 
     @Test
-    fun `historical duplicate printable ASCII workaround is preserved`() {
+    fun `genuine repeated printable ASCII is preserved`() {
         val input = listOf(
             'a'.code.toByte(),
             'a'.code.toByte(),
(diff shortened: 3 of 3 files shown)


GlassOnTin (changes requested):
Reviewed both halves; the substantive comments are on GlassOnTin/termlib#2, since that is where the work is. Short version: the approach is right, and I have asked for a bound on the held-DEL repeat and a question about the `notifyImeSelection` change before it merges.

The Haven side of this is fine on its own. Making `normalizeCoalescedInput` a passthrough is the correct consequence of moving the echo suppression to where the callback identity exists, and the renamed test does fail if the `[X,X] → [X]` heuristic comes back. One small thing once the rest settles: a function that only calls `toByteArray()` is a seam with nothing left in it — either inline it at the call site or keep it and say in the KDoc that it is deliberately empty, so nobody re-adds a heuristic to it later.

The blocker is the `.gitmodules` change, and it cannot ship even temporarily:

```
-	url = https://github.com/GlassOnTin/termlib.git
+	url = https://github.com/binhdnguyen/termlib.git
```

F-Droid's build directories are sticky. `git submodule init` does not overwrite a URL that is already registered, so a `.gitmodules` repoint is invisible to their builder — and a commit that only exists on a fork then fails to fetch there while every one of our own builds stays green. That froze the F-Droid channel for a day in #525, and it reaches their side only through a release tag, so it is slow to undo.

Merge order that avoids it: termlib#2 lands on `GlassOnTin/termlib` first, then this PR points the submodule at that SHA with the URL unchanged. Nothing else here needs to move.

Thank you for the report-to-fix on this one, and for testing it against a real Vietnamese Gboard — that is the part I cannot do from here.

GlassOnTin:
Applied and shipped in `8abb6df8e` on main. This shows as closed rather than merged because the submodule half had to land in our fork first. I applied the Haven-side files directly instead of merging the branch with the fork URL in it.

Your termlib commit kept your authorship: `8a9a689`, now on `GlassOnTin/termlib` main, with my `1dc990d` on top adding the bound on the held-DEL takeover. The Haven-side commit is authored by me and credits you in the message. This is the wrong way round; I should have preserved your authorship on that half but did not think of it until it was pushed. Rewriting main during a release to fix a trailer is not worth it, so I am recording it here instead.

I did not verify the part that matters most. The Vietnamese Gboard round trip is yours; I have no Telex IME here to repeat it.

One thing your PR taught the repo. It repointed `.gitmodules` at your fork so CI could fetch the review SHA. Every job went green while building it. This meant the checkmarks vouching for the PR were produced by executing code nobody had read yet. A submodule here is not vendored data: `settings.gradle.kts` pulls several in with `includeBuild()`, so their `build.gradle.kts` files run on the runner. Pinning the SHA does keep the content honest. Your branch was one commit touching two Kotlin files, so nothing was wrong with it. However, the bound came from me reading the diff rather than from anything enforcing it. CI now refuses a submodule URL outside our org and says what to do instead. The next contributor gets told in seconds rather than after a review round trip. That is a gap in our setup you surfaced, not a mark against the PR.

Thank you for this one, and for testing it on real hardware.

binhdnguyen:
> Reviewed both halves; the substantive comments are on GlassOnTin/termlib#2, since that is where the work is. Short version: the approach is right, and I have asked for a bound on the held-DEL repeat and a question about the `notifyImeSelection` change before it merges.
> 
> The Haven side of this is fine on its own. Making `normalizeCoalescedInput` a passthrough is the correct consequence of moving the echo suppression to where the callback identity exists, and the renamed test does fail if the `[X,X] → [X]` heuristic comes back. One small thing once the rest settles: a function that only calls `toByteArray()` is a seam with nothing left in it — either inline it at the call site or keep it and say in the KDoc that it is deliberately empty, so nobody re-adds a heuristic to it later.
> 
> The blocker is the `.gitmodules` change, and it cannot ship even temporarily:
> 
> ```
> -	url = https://github.com/GlassOnTin/termlib.git
> +	url = https://github.com/binhdnguyen/termlib.git
> ```
> 
> F-Droid's build directories are sticky. `git submodule init` does not overwrite a URL that is already registered, so a `.gitmodules` repoint is invisible to their builder — and a commit that only exists on a fork then fails to fetch there while every one of our own builds stays green. That froze the F-Droid channel for a day in #525, and it reaches their side only through a release tag, so it is slow to undo.
> 
> Merge order that avoids it: termlib#2 lands on `GlassOnTin/termlib` first, then this PR points the submodule at that SHA with the URL unchanged. Nothing else here needs to move.
> 
> Thank you for the report-to-fix on this one, and for testing it against a real Vietnamese Gboard — that is the part I cannot do from here.
> 

On `notifyImeSelection:` the symptom was Gboard Vietnamese Telex composing against glyphs that backspace had already removed from the shell. Haven SSH is non-full-editor (Secure). DEL was reaching the PTY but not shrinking the InputConnection Editable, so getTextBeforeCursor still returned the deleted text.

The content fix is `onUserDelKey()` shrinking the Editable. notifyImeSelection is the IME half of that: without updateSelection, Gboard kept its composing context even after the buffer had changed. The old `if (!fullEditor) return` meant that call was a no-op in Secure, so the shrink never reached the IME. Relaxing it to `if (!fullEditor && ed.isEmpty()) return` was for that path, not for a second bug. updateSelection only reports cursor indices, not buffer contents.

I did not A/B that line separately from the Editable shrink, and I did not retry the same sequence on Samsung / SwiftKey / AOSP LatinIME. If something else misbehaves after this, that is the first place I would look too.

The 5s bound and the focus/window cancel in 1dc990d are the right failure mode. Device testing here never hit a stuck takeover; the case we hit was Gboard going silent mid-hold, which the takeover is for.

binhdnguyen:
> Applied and shipped in `8abb6df8e` on main. This shows as closed rather than merged because the submodule half had to land in our fork first. I applied the Haven-side files directly instead of merging the branch with the fork URL in it.
> 
> Your termlib commit kept your authorship: `8a9a689`, now on `GlassOnTin/termlib` main, with my `1dc990d` on top adding the bound on the held-DEL takeover. The Haven-side commit is authored by me and credits you in the message. This is the wrong way round; I should have preserved your authorship on that half but did not think of it until it was pushed. Rewriting main during a release to fix a trailer is not worth it, so I am recording it here instead.
> 
> I did not verify the part that matters most. The Vietnamese Gboard round trip is yours; I have no Telex IME here to repeat it.
> 
> One thing your PR taught the repo. It repointed `.gitmodules` at your fork so CI could fetch the review SHA. Every job went green while building it. This meant the checkmarks vouching for the PR were produced by executing code nobody had read yet. A submodule here is not vendored data: `settings.gradle.kts` pulls several in with `includeBuild()`, so their `build.gradle.kts` files run on the runner. Pinning the SHA does keep the content honest. Your branch was one commit touching two Kotlin files, so nothing was wrong with it. However, the bound came from me reading the diff rather than from anything enforcing it. CI now refuses a submodule URL outside our org and says what to do instead. The next contributor gets told in seconds rather than after a review round trip. That is a gap in our setup you surfaced, not a mark against the PR.
> 
> Thank you for this one, and for testing it on real hardware.
> 

Ack on the merge path. I did not know about the sticky F-Droid submodule URL from #525; pointing .gitmodules at the fork so CI could fetch the review SHA was the wrong way to do that. The org-only URL check is the instruction I needed.

Authorship on the Haven half is fine as recorded. The Vietnamese Gboard round trip still holds on the build I tested: genuine `aa`, Telex composition after deleting `độc`, and holding Backspace through a syllable all behaved correctly. I have not re-run those checks specifically on `8abb6df8e + 1dc990d` yet, but the Haven-side changes are the same ones I tested, and `1dc990d` only adds a bound to the held-DEL takeover rather than changing the normal composition path.

Thanks for reviewing it closely and for tightening the submodule CI path as a result.

PR #590, opened 2026-08-25
fix(navigation): restore terminal after unlock

tsubus:
Keeps you on the Terminal tab after unlocking Haven, instead of sending you back to Connections.

 If Haven was fully restarted, it safely opens Connections because the old terminal session is no longer available.

### DIFF
--- a/app/src/main/kotlin/sh/haven/app/navigation/HavenNavHost.kt
+++ b/app/src/main/kotlin/sh/haven/app/navigation/HavenNavHost.kt
@@ -245,6 +245,10 @@ fun HavenNavHost(
     val navStateViewModel: NavStateViewModel = hiltViewModel()
     val selectedScreen by navStateViewModel.selectedScreen.collectAsState()
     fun requestScreen(screen: Screen) { navStateViewModel.select(screen) }
+    // The fresh pager starts at page zero when the host re-enters composition.
+    // Do not let that initial settled page overwrite the retained selection
+    // before the render effect below has restored it.
+    var restoreDone by remember { mutableStateOf(false) }
     // True while a manual page-swipe (pagerSwipeOverride, on the Terminal/SFTP
     // pages) owns the pager. While set, the settledPage→selection sync below is
     // suppressed: the override adopts the settled page explicitly on release, so
@@ -279,6 +283,7 @@ fun HavenNavHost(
         if (idx >= 0 && idx != pagerState.currentPage) {
             pagerState.scrollToPage(idx)
         }
+        restoreDone = true
     }
     // Manual swipes feed the source of truth: when the pager *settles* on a
     // page, adopt that screen as the selection. Keyed on settledPage only
@@ -289,7 +294,7 @@ fun HavenNavHost(
         // the settled page (it calls requestScreen on release); skip here so a
         // transient settledPage can't flip the selection mid-gesture. The
         // built-in-pager pages never set this flag, so they still sync here.
-        if (isUserSwiping) return@LaunchedEffect
+        if (!restoreDone || isUserSwiping) return@LaunchedEffect
         screens.getOrNull(pagerState.settledPage)?.let { settled ->
             if (settled != selectedScreen) {
                 requestScreen(settled)


GlassOnTin:
Thanks for this — the diagnosis in the first half is right, and it is a bug I
had not spotted. The fresh pager does settle on page zero when the host
re-enters composition, and the settled-page effect adopts that before the
restore scroll lands. Something has to hold it off.

I do not want to merge it as it stands, for two reasons.

**The latch also switches off the only path that recovers a stale selection.**

`selectionApplied` is set to `idx >= 0`, and `idx < 0` is a normal state, not
an error. The effect's own comment lists the cases: a tab still gated off, a
deep link to Desktop before its first session. The plainest one is Mail, which
`screens` includes only while a mail session is CONNECTED and drops on
disconnect. Disconnect while you are on the Mail tab and `selectedScreen` is
Mail while `screens` has no Mail, so `idx` is -1.

Nothing else in the file resets `selectedScreen` when the selected screen
disappears — I checked every use of it. The settled-page effect at line 288 is
the whole recovery: the pager clamps to a valid page, settles, and the sync
adopts it. With this patch that effect returns early in exactly that state, so
the selection stays pointing at a screen that is not there, the nav bar
highlights nothing, and swiping never fixes it. You would have to tap a tab.

A one-way latch avoids that, because what you need to know is "has the first
restore run", not "is the selection currently present":

```kotlin
var restoreDone by remember { mutableStateOf(false) }

LaunchedEffect(selectedScreen, screens) {
    val idx = screens.indexOf(selectedScreen)
    if (idx >= 0 && idx != pagerState.currentPage) {
        pagerState.scrollToPage(idx)
    }
    restoreDone = true
}

LaunchedEffect(pagerState.settledPage) {
    if (!restoreDone || isUserSwiping) return@LaunchedEffect
    ...
}
```

Set unconditionally, and left out of the effect's key list. Keeping it out
matters: the existing comment says the effect is keyed on `settledPage` only
so that a pending `requestScreen` to a not-yet-present tab is never clobbered
by a screens-list change, and adding a second key walks that back. The
ordering still works — `LaunchedEffect`s start in declaration order, the
restore effect suspends at `scrollToPage` with `restoreDone` still false, the
settled-page effect sees false and returns, and the scroll completing changes
`settledPage`, which re-runs it once with the correct page.

**The Terminal-to-Connections rewrite is doing the wrong thing.**

The premise is that a Terminal route is stale after process recreation because
the session is gone. The Terminal tab is not gated on sessions. `screens` has
`Screen.Terminal -> hasTerminalProfiles`, and that is `connections.any {
it.isTerminal }` — saved profiles, which are in the database and come back.
So after process death the tab is there, and the page renders
`EmptyTerminalState` with the connect prompt, swipeable by design. There is no
stale route to avoid.

It also contradicts a deliberate decision. The comment above
`navStateViewModel` says the `SavedStateHandle` is what carries the selection
across process death, and this makes that true for every route except
Terminal. Someone who lives on the Terminal tab would be sent to Connections
every cold start, which is close to the complaint your first half fixes.

I would drop that hunk and its test. If it turns out a restored Terminal route
does cause a problem, that wants its own issue with the symptom in it, because
the fix would probably belong in the terminal screen rather than in the
navigation state.

Happy to take an updated version. The first half with a one-way latch would go
in as-is, and it is a good catch.

GlassOnTin:
Merged as `fd256fec8`, rebased so your authorship is on the commit.

You had already made both changes before I got back to it, and the revised version is correct. I re-checked the ordering rather than just matching it against what I had suggested. When `idx < 0`, the restore effect falls through to `restoreDone = true` without scrolling. Consequently, the settled-page effect still runs and still adopts the pager's clamped page. That was the property I was worried about losing, and it survives.

The diagnosis was the part I had missed entirely. A fresh pager settling on page zero before the restore scroll lands is not something I would have found from reading the file.

One thing that is not covered, and that I am not going to pretend is: there is no test for the latch. `:app` has no JVM Compose test setup, only `NavStateViewModelTest`. Adding one means new test dependencies and a dependency-verification regen, which is not something to start while a release is building. So the ordering rests on your device testing and my reading of it. If the tab ever comes back wrong after an unlock, that latch is the first thing to look at.

Thanks for the revision, and for taking the review as read rather than arguing it.

PR #612, opened 2026-08-30
Add Serbian localization strings

eevan78:
Here is the initial submition of the Serbian (sr) translation.


GlassOnTin:
Thanks for this — Serbian is a good addition. Two things to fix before I can merge:

1. `app_name` — please keep it as `Haven`. Every other locale (ru, de, ja, …) keeps the app name untranslated, and the name on F-Droid and the Play Store is Haven.
2. `common_rename` is currently `Откључај`, which is the same as `common_unlock` (Unlock). Rename should be `Преименуј`.

70 of 290 strings is fine to merge — Android falls back to English for the missing keys, and the translation page will show the rest as open. If you'd like to keep going, `docs/i18n/strings.json` has the full list with context comments.

eevan78:
Thanks!
No problem, I'll continue with Haven, makes sense.

Point 2 was my mistake, I should have checked better.

I intend to continue and finalize the complete translation.

PR #615, opened 2026-08-31
feat(ssh): add Herdr session manager support

### LINKED ISSUE #611: Add Herdr as an optional Terminal/session manager alongside tmux, zellij, screen, and byobu
# Motivation

Herdr (https://herdr.dev/) is a third-party, agent-aware persistent terminal runtime and multiplexer. Its background server owns PTYs, processes, panes, layouts, workspaces, and named sessions, allowing terminal clients to detach and reconnect without stopping the work.

Haven can already connect to a host running Herdr, but the user must attach manually after connecting. A first-class profile option would make reconnecting predictable and give Herdr the same session-picker and restore integration as Haven's existing session managers.

This is useful beyond any particular agent CLI. One concrete workflow is running Hermes Agent inside Herdr, but Herdr can also host ordinary shells, servers, tests, and other terminal agents.

# Possible integration

Herdr exposes machine-oriented commands suitable for a Haven adapter:

- Start or attach to a named session: `herdr --session <name>`
- List sessions: `herdr session list --json`
- Stop a session: `herdr session stop <name> --json`
- Delete a session: `herdr session delete <name> --json`

Possible Haven behavior:

- Detect whether herdr is installed on the remote host.
- List existing named sessions in Haven's session picker.
- Create or attach to the selected session after connecting.
- Remember and offer to reopen previously used Herdr sessions.
- Show a clear message when Herdr is unavailable.
- Define whether Haven's remove action stops the session or deletes its saved state.
- Keep the integration independent of any particular agent CLI.

Herdr does not currently document a named-session rename command, so rename support could be omitted or handled separately.

# Responsibility boundary

Herdr persists the terminal and process layer. It does not replace application-level session storage.

For example, Hermes Agent still owns its conversation history and resume state. Herdr preserves the running terminal environment, and its Herdr-maintained Hermes integration can report a resumable session identity for restoration.

# References

- Herdr persistence and remote access: https://herdr.dev/docs/persistence-remote/
- Herdr CLI reference: https://herdr.dev/docs/cli-reference/
- Herdr session-state behavior: https://herdr.dev/docs/session-state/
- Haven's existing session-manager registry: https://github.com/GlassHaven/Haven/blob/main/core/ssh/src/main/kotlin/sh/haven/core/ssh/SessionManager.kt

w3lld1:
## Summary

- I added Herdr to the saved-profile and SSH session-manager options.
- I attach through a guarded `herdr --session <name>` command and parse session names from Herdr's JSON output.
- I made session removal stop Herdr and delete its saved state, while leaving rename unsupported.
- I added focused coverage for command generation and JSON parsing, and I made `create_connection` accept Herdr.

## Validation

- I ran a focused Kotlin probe that compiled `SessionManager.kt` and exercised Herdr attach/remove commands plus valid and malformed JSON parsing.
- I ran `git diff --check`.

I could not start the Gradle test task locally because this runner reached its native process/thread limit, so I left the full module suite to CI.

Fixes #611


GlassOnTin:
Thanks — the shape is right and the guarded attach matches the other four. I ran it locally and it doesn't build yet: adding HERDR to the enums breaks two exhaustive `when` sites the PR doesn't touch.

```
e: feature/connections/.../ConnectionsViewModel.kt:5819 'when' expression must be exhaustive (toSshSessionManager)
e: feature/terminal/.../SaveConnectionFromSession.kt:42 'when' expression must be exhaustive (pinRemoteCommand)
```

Adding `HERDR -> SessionManager.HERDR` to the first and `SessionManager.HERDR -> "herdr --session $name"` to the second makes `:app:compileArm64FullDebugKotlin` pass.

Two more gaps:

- The profile editor's session-manager dropdown (`ConnectionEditDialog.kt`, the `sessionManagerOptions` list around line 2457) still lists only tmux/zellij/screen/byobu, so HERDR is unreachable from the UI. The Settings screen iterates the enum entries and picks it up automatically.
- The JSON parser: the outer regex is greedy (`.*` up to the last `]`), so the capture runs to the end of the document and the name regex then matches every `"name"` in it. If session objects nest workspaces/tabs/panes (the CLI reference has rename commands at all three levels), their names leak into the picker: `{"sessions":[{"name":"work","workspaces":[{"tabs":[{"name":"shell-1"}]}]}]}` parses as `work, shell-1`. The docs don't publish the list shape, so I'd parse with a real JSON parser rather than regex, or at least anchor to the top-level array.

Minor: the other four labels are lowercase (tmux, zellij, screen, byobu); HERDR uses "Herdr".

The 5 new tests pass locally (`:core:ssh:testDebugUnitTest`, 5/5).

GlassOnTin:
I pushed a follow-up commit onto the branch with the fixes from my review (maintainer edits were enabled — apologies if that wasn't wanted, easy to revert).

- Added the two missing `HERDR` branches (`ConnectionsViewModel.toSshSessionManager`, `SaveConnectionFromSession.pinRemoteCommand`) — `:app:compileArm64FullDebugKotlin` now passes.
- Herdr is now in the profile editor dropdown, and the label is lowercase `herdr` to match the others.
- The list parser now uses `org.json` instead of the greedy regex. I checked the actual shape against the herdrdev/herdr source (`src/session.rs` `SessionInfo`): `herdr session list --json` prints a flat `{"sessions":[{"name","default","running","socket_path","session_dir"}]}` — no nesting today, so the leak I flagged couldn't fire, but the regex would break the day the shape grows. The parser now mirrors Herdr's own `validate_name` (ASCII alnum `.` `_` `-`, max 64).
- Bonus: wired the detach keybinding (`Ctrl+B q` — Herdr's default prefix+detach, from `config/model.rs`), which the other four managers have.
- Three more tests (real SessionInfo shape, nested-name isolation, invalid-name filtering). `:core:ssh:testDebugUnitTest` 8/8, plus the connections and terminal module suites, green locally.

GlassOnTin:
Reviewed. The change is well-scoped and matches how the other multiplexers are wired: one `HERDR` entry in each of the two enums, the attacher mapping, the MCP `create_connection` allow-list, and the JSON session-list parse with name validation mirroring Herdr's own `validate_name`. The `command -v herdr` guard in the attach command is the right call so a profile pointing at a host without Herdr fails with a readable message instead of a dead tab. Tests cover command generation, the real `SessionInfo` shape, nested-name non-leakage, and invalid-name skipping.

Two notes, neither blocking:

- `core/ssh/build.gradle.kts` adds `org.json:json:20260814` as a test dependency. That version is already pinned in `gradle/verification-metadata.xml` and used by other test modules, so dependency verification will pass as-is; no metadata regen needed.
- I have not approved the paused CI run: this PR touches a Gradle build script, and first-time-contributor workflow approvals for build-script changes wait for a maintainer. CI will run once a maintainer approves.

Holding for maintainer merge.

PR #616, opened 2026-09-01
Update Serbian translation

eevan78:
Here is another chunk of strings translated to Serbian.


GlassOnTin:
Thanks for keeping the Serbian strings up to date, and for the app-name fix (Лука -> Haven).

One string will crash at runtime: `agent_summary_uploaded_bytes_via` ends with `%2$` instead of `%2$s`:

```
отпремљено је %1$d бајтова преко %2$
```

The code passes two arguments (`getString(R.string.agent_summary_uploaded_bytes_via, n, backend)`), and an incomplete `%2$` specifier makes `String.format` throw `UnknownFormatConversionException` as soon as that audit line renders. Adding the missing `s` fixes it:

```
отпремљено је %1$d бајтова преко %2$s
```

Everything else checks out: all four files are valid XML, every string name exists in the English source, and the other format placeholders match. Fix that one line and this is good to merge.

eevan78:
Here is the fix...

I'll continue translation work.

GlassOnTin:
The placeholder fix looks right, and the CI run on this push is in flight. Once it's green this is good to merge.

eevan78:
It still fails the check translation export is current... I don't see why.

GlassOnTin:
Quick note on the CI failure and what I changed.

Your fix was correct, but the run still failed on a different gate: "Check translation export is current". That gate regenerates `docs/i18n/strings.json` (the file behind the web translation page) and fails if it doesn't match the `strings.xml` sources. Adding the new Serbian strings changed the sources, so the export needed regenerating to match.

Since your PR allows maintainer edits, I ran `python3 scripts/i18n_export.py` and pushed the regenerated file to your branch (85e9f432). The diff is only your Serbian strings appearing in the export, nothing else. CI is re-running now; once it's green this is good to merge.

For future translation PRs, that's the one extra step: run `python3 scripts/i18n_export.py` after editing any `strings.xml` and commit the result. Thanks again for keeping Serbian up to date.

PR #619, opened 2026-09-02
Update of the Serbian translation

eevan78:
Here is another batch of new strings.
It also fixes some typos in the previous commits, and adds proper Serbian label to the translation page.


GlassOnTin:
Thanks — the strings themselves are clean: I diffed every added/changed `values-sr` string against the English source and all names exist and all format specifiers match (the placeholder bug that caught the last batch isn't here), and you regenerated `docs/i18n/strings.json` so the export gate passes.

One thing CI won't catch, in `docs/i18n/translate.js` line 14: the Serbian label was added under a second `ru:` key instead of `sr:`:

```js
ru: "Russian · Русский", ru: "Serbian · српски",
```

In a JS object literal the duplicate key silently overwrites the first, so the picker would show "Serbian · српски" in Russian's slot and Serbian (the `sr` option `translate.html` now adds) would have no label. One-character fix:

```js
ru: "Russian · Русский", sr: "Serbian · српски",
```

Once that's in, this is good to go from my side.

GlassOnTin:
Confirmed — `translate.js` line 14 now reads `sr:` and CI is green on the new head. The strings and the export are all clean, so this is good to go from my side. Thanks for the quick turnaround.

PR #624, opened 2026-09-04
Update of the Serbian translation

eevan78:
Some more finished modules, and update app module translation (together with some small translation fixes).


GlassOnTin:
Thanks for the translation work. One placeholder bug to fix before this can merge:

`app_desktop_add_foreign_distro_confirm_body` uses `%1$d` for the size, but the English source is `"This downloads and extracts the %1$s rootfs for emulated use, about %2$d MB."` The size argument is `%2$d`. With `%1$d`, `getString` passes the rootfs name (a String) to a `%d` conversion and throws `IllegalFormatConversionException` at runtime, so the add-foreign-distro confirm dialog would crash in Serbian.

Change `отприлике %1$d MB` to `отприлике %2$d MB`.

I checked the other 377 added strings for placeholder mismatches against the English sources; this is the only one.

PR #627, opened 2026-09-07
add(themeing): various new themes

HolgerHuo:


GlassOnTin:
This is in v5.87.78 — six new color schemes in the terminal color picker, with scheme selections persisted by name so existing settings carry over. GitHub release is up; F-Droid build follows on their cadence.

Issue #245, opened 2026-06-12
Toolbar issues


jeyjai:
<img width="1080" height="232" alt="Image" src="https://github.com/user-attachments/assets/7ff56810-6427-48b1-a680-f2c17ee44c13" />

If you disable the Voice key you get this space under the attach key but it can't be filled by any other keys, it stays an empty box. And if you disable the attach key at the same time you get an empty box on both rows. Just a wide gap you can't use for any other keys. Its basically like they're still there but hidden.

-

It would also be nice if we could disable the key for VNC Desktop. I personally dont have any use for it so I'll rather use that space for something i do use. Even being able to move it to the opposite side so its out of view without moving other buttons at the same time would be helpful.

-

Last issue. If you add or remove a key in the toolbar settings after reorganizing your keys they all get returned to their default positions, which i think is the order they're listed in the toolbar settings menu.

I was even able to reproduce this several times in a row by simply pressing the pen key to edit the toolbar when i already had reorganized the keys, but im no longer able to reproduce this and ive changed a loy since then so not really sure what caused it to happen or why it stopped.

GlassOnTin:
Thanks — three separate things here, all fair. Taking them in turn:

**1. Empty box when you disable Voice (or Attach).**
Confirmed. The bottom of the toolbar is an *aligned* two-row grid where the row-1 and row-2 keys are paired column-by-column — Voice is deliberately stacked directly under Attach so they line up. When you turn one of a pair off, the grid keeps the column and draws an empty ~32 dp cell instead of sliding the neighbouring keys up to fill it. That's the "still there but hidden" gap you're seeing. Fix: collapse a column once both of its cells are empty (and let a lone remaining key reflow), so a disabled key actually gives its space back. I want to device-test this one because the column-pairing is intentional (it's what keeps Attach/Voice aligned), so I don't want the fix to misalign the rest.

**2. Disable / move the VNC Desktop key.**
Confirmed there's no way to do this today — unlike the other keys, the desktop button isn't a configurable toolbar key; it's a fixed control that's shown automatically whenever the session can open a desktop, pinned to the bottom-left. Plan: add a setting to hide it, and — matching the existing "edit-mode controls placement" option — let you move it to the opposite side so it's out of the way without disturbing your other buttons.

**3. Keys reset to defaults after adding/removing a key.**
[...]

PR #628, opened 2026-09-07
fix: desktop/voice button position

HolgerHuo:
The keyboard toggle and the auto-shown Desktop (VNC/RDP) key were pinned to a fixed column, so hiding the Desktop key left an empty dead cell under the keyboard toggle, and the keyboard key could not be moved. 

This bug was mentioned in https://github.com/GlassHaven/Haven/issues/245 but seems to have been reintroduced.

### DIFF
--- a/core/toolbar/src/main/kotlin/sh/haven/core/toolbar/KeyboardToolbar.kt
+++ b/core/toolbar/src/main/kotlin/sh/haven/core/toolbar/KeyboardToolbar.kt
@@ -532,26 +532,19 @@ private fun AlignedToolbarContent(
     val (row1Left, row1Right) = splitAroundNav(layout.row1)
     val (row2Left, row2Right) = splitAroundNav(layout.row2)
 
-    // Pin the leading icon keys to a fixed 2-column grid so the Attach
-    // (row 1) and Voice/secure-keyboard (row 2) icons always stack at
-    // column 2 — col 1 holds the keyboard toggle (row 1) / VNC-desktop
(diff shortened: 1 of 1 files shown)


GlassOnTin:
Reviewed the diff. The diagnosis matches #245: the keyboard toggle is pinned to a fixed leading column with the desktop key under it, so with the desktop key hidden you get a dead cell and the keyboard key ignores its saved position.

Letting the keyboard key flow as a normal item and hanging the LEFT-placed desktop key beside it covers the cases I could trace: keyboard in either row, keyboard moved to the trailing segment (desktop falls back to the trailing edge), and keyboard absent entirely. The edit-mode rework (done button as a fixed cell, keyboard no longer special-cased in drag) is consistent with the new model.

Two notes, neither blocking:

- LEFT placement now renders the desktop key in the top row beside the keyboard instead of below it. That is a visible change for existing layouts even when nothing is disabled - worth calling out in the changelog.
- Edit mode still pins Attach to the front of row 1 unconditionally, while the live render only pins Attach/Voice as a pair. Pre-existing asymmetry, just noting it since this diff touches the same code.

CI is green including the compose-heavy test job. This is a behavior change to the #245 layout contract, so it will sit for a maintainer merge decision rather than auto-merging.

GlassOnTin:
This is in v5.87.78 — the keyboard toggle and the Desktop key are ordinary toolbar items now, so hiding the Desktop key no longer leaves the dead cell, and a left-placed Desktop key sits beside the keyboard in the top row. GitHub release is up; F-Droid build follows on their cadence.

PR #629, opened 2026-09-07
feat(ui): togglable nav bar items

HolgerHuo:
Navbar items' auto hide no longer works. This PR introduces per item configuration of item display in navbar.


GlassOnTin:
Reviewed the full diff. Clean and well-structured:

- The tab filter is now a pure `visibleScreens()` function with real unit tests covering AUTO/SHOW/HIDE, the always-visible guard, and ordering — good move.
- The DataStore layer is tested too, including the legacy `always_show_all_tabs` read-time fallback and the case where saving an all-AUTO map disables the fallback again.
- Atomic single-edit commit of order + visibility (avoids the pager re-rendering twice) and the draft-map-in-dialog approach that avoids disposing the Settings page mid-edit are both right.
- Persistence is by route string with `fromName` falling back to AUTO, so enum changes can't corrupt saved state.
- CI is green on the branch; nothing touches build scripts, workflows, or natives.

Two nits, neither blocking:
1. The Serbian strings for `settings_navigation_tabs_title`/`_subtitle` still carry the old "Screen order" wording (Редослед на екрану / Мењање редоследа картица за навигацију на дну) — now mistranslated for the new meaning. Worth a follow-up in the next translation pass.
2. `screenRoutesForcedVisibleByLegacy()` hard-codes the five routes in core/data, away from `Screen` — necessary given the layering, but it silently drifts if a screen route is ever renamed. A comment cross-referencing Screen.kt already covers this; just flagging it.

Holding for maintainer merge per policy (code change), but I have no concerns.

GlassOnTin:
This is in v5.87.78 — each navigation tab has an Auto / Show / Hide choice, replacing the single show-all switch. An existing show-all preference carries over as Show. GitHub release is up; F-Droid build follows on their cadence.

